Issue #6, opened 2017-05-10
Repo structure & .gitignore & README


Drakulix:
Create some folders.

Maybe
  - resources
  - frontend
    - .gitignore
  - backend
    - .gitignore
  - README.md
  - .gitignore
  - bla bla bla

markwitz:
maybe a tools/scripts folder containging usefull shellscripts for testing or setting up sample data?

Drakulix:
Maybe as a subfolder of resources? At least that folder is what I wanted to use for sample data. Or maybe tools could be for code and resources for data?

markwitz:
Sounds good

Drakulix:
Closed in favor of #78

PR #10, opened 2017-05-14
dockerfile, first db creation, search example

henderkes:


Dockerfile
```diff
@@ -0,0 +1,38 @@
+# escape=`
+
+RUN pip install pymongo
+RUN pip install Flask
+
+FROM microsoft/windowsservercore as msi
```
Drakulix: Why are we using a windows image as a base? Does this even work outside of windows?

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN pip install Flask
+
+FROM microsoft/windowsservercore as msi
+
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
+
+ENV MONGO_VERSION 3.4.4
+ENV MONGO_DOWNLOAD_URL http://downloads.mongodb.org/win32/mongodb-win32-x86_64-2008plus-ssl-${MONGO_VERSION}-signed.msi
+ENV MONGO_DOWNLOAD_SHA256 73d1121a1fe58fe29c4fb7b6f4b79a4d33afa666f357960419ef83635396f977
+
```
Drakulix: There is already a mongo docker image, no need to replicate it https://hub.docker.com/_/mongo/

Drakulix (changes requested):
We probably only need the `Dockerfile` for our own python webserver (that container should be based on the official [`python` docker image](https://hub.docker.com/_/python/) and the rest of the services in a [`docker-compose.yaml`](https://docs.docker.com/compose/).

The `first.py` is a good start and helpful, but please use a proper name (maybe main.py) and create a folder for it. (See #6)

Dockerfile
```diff
@@ -0,0 +1,38 @@
+# escape=`
+
+RUN pip install pymongo
+RUN pip install Flask
+
```
Drakulix:
```
Step 1/22 : RUN pip install pymongo
Please provide a source image with `from` prior to run
```

Cannot even test it..

henderkes:
updated

LICENSE
```diff
@@ -1,21 +0,0 @@
-MIT License
```
Drakulix: ähm, the license is not an "unnecessary file".
Your first commit does not seem right.
Please try to do the following steps:

```
git checkout master
git pull
git checkout DubbleClick-patch-1
git rebase -i master
```

And remove your [last commit](6cf8cd0bd00f9ca61d1350be19386f2fb7e2bb1b) (skip) from the rebase.
Also please skip all commits, that are already part of master (all travis related commits) and then fix any remaining conflicts, if any (should all be fine).

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3'
+
+services:
+    mongodb-service:
+        build: ./mongodb
```
Drakulix: This is unnecessary. You can just do `image: mongo`, that's it.

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+        ports:
+            - 27017:27017
```
Drakulix: Unnecessary, please see the next comment.

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+    python-file:
+        build: ./pythonfile
+        ports:
+            - 8000:27017
```
Drakulix: Please just do:

```
links:
    - mongodb-service
```
to automatically route all mongodb ports to only the python container and just open the port of the outgoing webserver port of flask.

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+        depends_on:
+            - mongodb-service
```
Drakulix: Please add a trailing new line

mongodb/Dockerfile
```diff
@@ -0,0 +1,25 @@
+# Pull base image.
```
Drakulix: File is unnecessary (see comment regarding `image:` key in docker-compose.yml)

pythonfile/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM python:3.6.1
```
Drakulix: Please use `3` instead of `3.6.1`.

pythonfile/requirements.txt
```diff
@@ -0,0 +1,2 @@
+flask
+pymongo
```
Drakulix, line 2: :+1:

pythonfile/Dockerfile
```diff
@@ -0,0 +1,2 @@
...
+CMD ["python", "first.py"]
```
Drakulix: Example taken from the python docker image [hub website](https://hub.docker.com/_/python/):

```
FROM python:3

WORKDIR /usr/src/app

COPY requirements.txt ./
RUN pip install --no-cache-dir -r requirements.txt

COPY . .

CMD [ "python", "./your-daemon-or-script.py" ]
```

By coping you can get the source files into the image (also fixing your windows problem / see slack) and redistribute our whole app as a container.
Also running `pip` will install all dependencies.

Drakulix (changes requested):
Thanks! Looks already very good. Just some nitpicks and smaller details.

Drakulix (dismissed):
Great!

You git history still seems a little confusing, although all files are here.
(I will fix that myself and merge asap)

Drakulix (approved):
@DubbleClick I have rebased your work. Can you check, that everything is still there?

Afterwards it's ready for merge.

PR #11, opened 2017-05-14
Initial Travis CI file

Krisselchen:
added initial Travis CI file with substructure for Python and Docker.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+sudo: required
+
+language: python
+
+python:
+  - "2.6"
+  - "2.7"
+  - "3.2"
+  - "3.3"
+  - "3.4"
+  - "3.5"
+  - "3.6"
+
+install: "pip install flask"
+
+services:
+  - docker
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ sudo: required
 language: python
 
 python:
-  - "2.6"
-  - "2.7"
   - "3.2"
   - "3.3"
   - "3.4"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,17 @@
+sudo: required
+
+language: python
+
+python:
+  - "2.6"
+  - "2.7"
```
Drakulix: I think we have explicitly stated not to support python 2, so you can remove these.

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+  - "3.2"
+  - "3.3"
+  - "3.4"
+  - "3.5"
+  - "3.6"
```
Drakulix, line 10: We probably also don't need to support every python3 version, but thats open to discussion.

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+
+install: "pip install flask"
+
+services:
+  - docker
```
Drakulix, line 15: Please add any `script`-key to make travis pass. (`true` should be enough as a command)

Drakulix (changes requested):
Thanks! I have requested some smaller changes. Otherwise this is good for merge.

.travis.yml
Krisselchen: I was not sure about it. Therefore I also took those. It is reduced to 3.x now.

.travis.yml
Krisselchen: Done. Travis is passing now.

Drakulix (approved):
Looks good!

PR #13, opened 2017-05-21
Docker elasticsearch extension

henderkes:
first integration with example usage of elasticsearch, opening localhost:9200 for web access and elasticsearch:9200 for usage within docker

### DIFF
--- a/pythonfile/first.py
+++ b/pythonfile/first.py
@@ -1,4 +1,9 @@
 from pymongo import MongoClient
+import elastic
+import time
+
+time.sleep(5) #wait for initialisation of elasticsearch and mongodb...
+#FIXME: exception handling
 
 jsonstring = {
     "title": "test_project",
@@ -83,3 +88,4 @@ if __name__ == "__main__":
     for project in res:
         print(project)
         print()
+    elastic.elastic_example()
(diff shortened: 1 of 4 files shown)


pythonfile/first.py
```diff
@@ -1,4 +1,9 @@
 from pymongo import MongoClient
+import elastic
+import time
+
+time.sleep(5) #wait for initialisation of elasticsearch and mongodb...
```
Drakulix, line 5: Okay for testing, but this cannot be the solution in the long term.
We will have to wait for the sockets to become available.
Please add another `FIXME` here, so we don't forget that.

Drakulix (changes requested):
Great work.

pythonfile/first.py
henderkes: there is a FIXME right under that in the next line, or am I missing something? 80 character limit...
edit: to clarify, the initialisation issue will (have to?) be fixed with exception handling

Drakulix: > edit: to clarify, the initialisation issue will (have to?) be fixed with exception handling

So thats the same FIXME then. Good to go in that case.

Drakulix (approved):

PR #15, opened 2017-05-23
Flask api

svirovsk:
The api has the three main endpoints  (add_project, get_project_by_id and search) plus delete_project and index, which returns the first project imported via create_project().
I've tried to test the endpoints with curl. index(), get_project_by_id(id) and delete_project() work. The request for add_project() is kind of long (the whole jsonstring) and I haven't tested it yet, but it's more or less the same as in create_project() and that one works. It adds date_created and date_published to the jsonstring, validates it and adds an uuid at the end. The functions don't handle errors yet.
Unfortunatelly, I can't run the docker container with elasticsearch, so it'll be nice if someone of you runs it and tells me if it works. Here, I'm not sure again how to design the curl request.
I haven't run the container with all changes and there might be other errors as well.


pythonfile/Dockerfile
```diff
@@ -7,4 +7,4 @@ RUN pip install --no-cache-dir -r requirements.txt
 
 COPY . .
 
-CMD [ "python", "./first.py" ]
\ No newline at end of file
+CMD [ "python", "./first.py", "./flask-api.py" ]
```
Drakulix: Was this tested? Does this work that way?

pythonfile/flask_api.py
```diff
@@ -0,0 +1,87 @@
...
+
+app=Flask(__name__)
+
+
+
+_id=uuid.uuid4()
+
+
+# add project to test with
+@app.before_first_request
```
Drakulix: Please remove this `^M` character

pythonfile/flask_api.py
```diff
@@ -0,0 +1,87 @@
...
+def create_project():
```
Drakulix: Is this some kind of test data?

pythonfile/flask_api.py
```diff
@@ -0,0 +1,87 @@
...
+    r = validator.validate_manifest(jsonstring)
+
+    if r == None:
+        
+        jsonstring['_id']=_id
+        coll.insert(jsonstring)
+        elastic.store_json("test", "projects", jsonstring)
+
+    
+@app.route('/', methods=['GET','POST'])
```
Drakulix: Also for testing?

pythonfile/flask_api.py
```diff
@@ -0,0 +1,87 @@
...
+def index():
+    return jsonify(coll.find_one(_id))
+
+# receive manifest as a jsonstring
+# {'manifest':manifest}
+@app.route('/add_project')
```
Drakulix: What happens if I do a `GET` request here? Should this be `POST` only?

pythonfile/flask_api.py
```diff
@@ -0,0 +1,87 @@
...
+def add_project():
+
+    manifest=request.json['manifest']
+    manifest['date_creation']=time.strftime("%Y-%m-%d")
+    manifest['date_update']=time.strftime("%Y-%m-%d")
+    json5.dumps(manifest)
+    schema = open("manifest_schema.json")
+    validator = ManifestValidator(schema)
+    r = validator.validate_manifest(manifest)
```
Drakulix: What is `r`? If it can contain an error, please write it to `stderr` for the moment.

pythonfile/flask_api.py
```diff
@@ -0,0 +1,87 @@
...
+
+    if r == None:
+        _id=uuid.uuid4()    
+        manifest['_id']=_id
+        coll.insert(manifest)
+        elastic.store_json("test", "projects", manifest)
+        return ('project imported \n')
+    
+    
+@app.route('/get_project_by_id/<uuid:project_id>', methods=['GET','POST'])
```
Drakulix: This should be `GET` only.

pythonfile/flask_api.py
```diff
@@ -0,0 +1,87 @@
...
+def get_project_by_id(project_id):
+    res=coll.find_one({'_id':project_id})
+    return jsonify(res)
+
+@app.route('/delete_project/<uuid:project_id>')
```
Drakulix: Again what method? I assume `POST` or `DELETE`?

pythonfile/manifest_validator.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, json_schema):
+        self.schema = json5.load(json_schema)
+        self.validator = Draft4Validator(self.schema, format_checker=FormatChecker())
+
+    def validate_manifest(self, json_manifest):
+        try:
+            manifest = json5.load(json_manifest)
+            result = self.validator.validate(manifest)
+            return result
+        except FileNotFoundError:
```
Drakulix: How is this error possible, if `json_manifest` is a string? Can `json5.load()` also accept filenames?
Or is this a left-over?

Drakulix (changes requested):
Generally already very good, I have requested some smaller changes.

Also we should try to design a RESTful API. While this is okay and supports all required use cases, I recommend to read [this article](http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#requirements), especially the very first paragraph (the rest does only partly apply to our current situation, but is still a rather good read).

I would suggest to change the api endpoints as follows:

```
GET /api/projects (list of projects - not required, but would be nice later, so I document that here)
POST /api/projects (upload manifest, return uuid of created project on success, error message else)
GET /api/projects/<id> (get project by id)
DELETE /api/projects/<id> (delete project by id)
POST /api/projects/search (elasticsearch endpoint)
```

The api prefix is suggested, because the frontend will most likely be served on the index. This way we can easily differentiate between the two.

Those are only suggestions and we can move them to separate issue, if time is short, they do not block the merge of this request (the requested changes do), although I would certainly like to see those rather soon then later.

pythonfile/flask_api.py
Drakulix: Or better write that to the response.body

pythonfile/flask_api.py
svirovsk: Well, I don't know how. I don't see it, when I try to edit it. Even if I write it again, it doesn't see the changes

pythonfile/flask_api.py
svirovsk: yes, it was supposed to be

pythonfile/flask_api.py
svirovsk: By default it just listens for GET requests, but of course, I can explicitly define the methods

pythonfile/flask_api.py
svirovsk: I suppose `DELETE` is fine, but what should the function return?

pythonfile/flask_api.py
```diff
@@ -54,29 +54,33 @@ def add_project():
...
+
+@app.route('/api/projects', methods=['GET'])  
+@app.route('/api/projects/<uuid:project_id>', methods=['GET'])
 def get_project_by_id(project_id):
     res=coll.find_one({'_id':project_id})
     return jsonify(res)
 
-@app.route('/delete_project/<uuid:project_id>')
+# returns DeleteResult.raw_result
+@app.route('/api/projects', methods=['DELETE'])
```
Drakulix: What happens if project_id is not given here?

pythonfile/flask_api.py
```diff
@@ -54,29 +54,33 @@ def add_project():
...
     return jsonify(res)
 
-@app.route('/delete_project/<uuid:project_id>')
+# returns DeleteResult.raw_result
+@app.route('/api/projects', methods=['DELETE'])  
+@app.route('/api/projects/<uuid:project_id>', methods=['DELETE'])
 def delete_project(project_id):
-    coll.delete_one({'_id':project_id})
-    return ('project deleted \n')
+    return jsonify(coll.delete_one({'_id':project_id}).raw_result)
```
Drakulix: jsonify? This does not seem right, I will search for a better solution

Drakulix (changes requested):
Great, much much better.

Drakulix (dismissed):

Drakulix (approved):
Updated

PR #16, opened 2017-05-23
Feature/top bar

Rintel:
# What? 

I've created a component for the top- and sidebar, added some styling, fixed some packages, added fonts and created a simple readme with an installation description.


Rintel:
Have a look, leave a comment or just approve it if you don't find anything suspicious.

PSDavidSt (commented):
looks good so far

suzanap:
there is a small problem with the side bar if you have more content on the page and you are obliged to scroll down. in this case, the side bar doesn't continue. but i don't think that it is a problem now

Rintel:
@suzanap I'm gonna dig into that issue later on, we have to create a base first where all of us can start properly. Therefore I need to refactor some stuff, but before that happens we really need every  ones task merged into master.

Rintel:
could some please approve this pull request?
@suzanap @PSDavidSt

suzanap (approved):

Issue #28, opened 2017-05-24
API Wiki


markwitz:
pease review:
[Wiki](https://github.com/Drakulix/knex/wiki/API)

markwitz:
how do we handle json5 input? 
Context-Type: application/json5 ?

Drakulix:
Yes JSON5 is handled by a special mime-type where accepted.

Overall I think the error codes are too simple. An internal server error for example is always possible. Other error types should be added where appropriate.

```
@app.route('/api/projects/add', methods=['PUT'])
onError: 404
```
Should be just /api/projects
Error code cannot be 404, the project does not exist.
Either `BadRequest` or `InternalServerError` (depending on the validity of the json(5).

```
@app.route('/api/projects/search', methods=['POST'])

{
  "query": {
  }
}
```

Why that query field there, is that actually ES-syntax? Just use the query object itself, no need to wrap it.

JamesComey:
Guys can we please only use GET and POST, I tried to do a delete query and couldn't get it working. I used HttpRequester(firefox addon) and it just didn't work... No clue if it is the addon or flask or whatever...

Lets just use
/api/projects/add
/api/projects/edit/id
/api/projects/delete/id
...
instead?  =) 
Not sure if, but I guess it will also be simpler for the frontend team... come on let's do this ^^

Drakulix:
no please use restful naming conventions. @JamesComey please try Postman and if that still does not work we can make an exception for the delete call only.

PR #46, opened 2017-05-27
Searchapi

JamesComey:
Working searchfunction with api/projects/dummyadd function, which adds a few example projects to es so the frontendteam can use the search.
Note: in docker port 9200 gets opened to elastic search, this is for test purposes and should be closed later on...


docker-compose.yml
```diff
@@ -6,6 +6,8 @@ services:
 
     elasticsearch:
         image: elasticsearch
+        ports:
+            - "9200:9200"
```
Drakulix, line 10: This looks like something that should be used temporarily for debugging and not on the public master branch. Please remove

pythonfile/flask_api.py
```diff
@@ -3,16 +3,18 @@
...
+from elasticsearch.exceptions import RequestError
 
 
 client=MongoClient('mongodb:27017')
 db=client.knexDBmh1
 coll=db.projects
 
-schema = open("manifest_schema.json")
-validator = ManifestValidator(schema)
+#validator = ManifestValidator(schema)
```
Drakulix, line 16: ? Please revert

pythonfile/flask_api.py
```diff
@@ -36,29 +39,103 @@ def add_project():
...
         coll.insert(manifest)
-        elastic.store_json("test", "projects", manifest)
+        #inserts in elasticsearch now
+        res = es.index(index="projects-index", doc_type='Project', id=manifest['id'], body=manifest)
         return make_response(manifest['id'])
     else:
         return make_response((error, '302'))
 
-@app.route('/api/projects/<uuid:project_id>', methods=['GET'])
+@app.route('/api/projects/get/<project_id>')
```
Drakulix, line 50: Why? please revert.

pythonfile/flask_api.py
```diff
@@ -36,29 +39,103 @@ def add_project():
...
 def get_project_by_id(project_id):
     res=coll.find_one({'_id':project_id})
     return jsonify(res)
 
-@app.route('/api/projects/<uuid:project_id>', methods=['DELETE'])
+
+@app.route('/api/projects/delete/<project_id>')
```
Drakulix, line 56: Same. See comment on #28

pythonfile/flask_api.py
```diff
@@ -36,29 +39,103 @@ def add_project():
...
 def delete_project(project_id):
-    if coll.delete_one({'_id':project_id}).deleted_count != 0:
+    #deletion from elastic search index
+    try:
+        res = es.delete(index="projects-index", doc_type='Project', id=project_id, refresh=True)
```
Drakulix, line 60: Please adjust. In case we got a DB inconsistency, we should still try to delete the MongoDB entry. If both fail, we should return an error.

pythonfile/flask_api.py
```diff
@@ -36,29 +39,103 @@ def add_project():
...
         return make_response('Success')
     else:
         return make_response('Project not found', 404)
 
-# receive body of elasticsearch query
+
+
+#Note, works but need better exception handlin on empty index
 @app.route('/api/projects/search', methods=['POST'])
+#@app.route('/api/projects/search')
```
Drakulix, line 75: ? Please remove that line

pythonfile/flask_api.py
```diff
@@ -36,29 +39,103 @@ def add_project():
...
 def search():
-    res=es.search(index="test", doc_type="projects", body=request.json)
-    return jsonify(res)
+    try:
+        res=es.search(index="projects-index", doc_type="Project", body=request.json)
+        return jsonify(res)
```
Drakulix, line 79: Could you add a TODO here?
I feel like the search result format is not very nice and we probably should filter it somehow or at least discuss this with the frontend team.

pythonfile/flask_api.py
```diff
@@ -36,29 +39,103 @@ def add_project():
...
+    except RequestError:
+        return ("Some Error")
+    except:
+       return ("Error: Index probablly empty!")
+
+# dummy add a few projects to es
+#Note: deleting these projects will throw not found error bcause they are in es only, but they are deleted
+@app.route('/api/projects/dummyadd', methods=['GET'])
```
Drakulix, line 87: This is nice for testing, but:
1. We already have example files in the repository
2. This will be unnecessary once the POST /api/projects call is working.

So while it is nice, that you got yourself a solution for testing, please remove (all of) this for merge.
I do not want debug code in my master. Even if we decide to use this for a demo, this can live in a separate branch.

Drakulix (changes requested):
Mostly good.
The Search API Endpoints look solid, as well as most of the additions for indexing and deletion to the rest.

Still I have requested some rather broad changes. Please read and fix or discuss.

Drakulix:
Also while a merge is technically fine, if you rebase your changes on current master, that would be even better. (See my git presentation or pm me).

PR #47, opened 2017-05-27
Flask file upload

henderkes:
single and multi file uploads are working as of now, exception handling will be improved later


Drakulix (changes requested):
Code-wise Awesome.

But something is fishy with your formatting (maybe line endings?)
I will not add a commit, that rewrites the whole dockerfile, flask_api.py and manifest_schema.json. And it is impossible to review any changes to the flask_api.py this way.

henderkes:
should be fixed now

backend/flask_api.py
```diff
@@ -0,0 +1,131 @@
...
+    res=elastic.es.search(index="test", doc_type="projects", body=request.json)
+    return jsonify(res)
+
+@app.route('/upload')
+def upload():
+    return render_template('upload.html')
+
+@app.route('/api/projects/upload', methods=['GET', 'POST'])
+def upload_file():
+    if request.method == 'POST':
```
Drakulix: Why does this even accept GET requests?

backend/flask_api.py
```diff
@@ -0,0 +1,131 @@
...
+        if 'file' not in request.files:
+            flash('No file part')
+            return redirect(request.url)
+        file = request.files['file']
+        if file.filename == '':
+            flash('No selected file')
+            return redirect(request.url)
+        securefilename = secure_filename(file.filename)
+        if file and uploader.allowed_file(securefilename):
+            file.save(os.path.join(app.config['UPLOAD_FOLDER'], securefilename))
```
Drakulix: 1. Frontend will mostly never use an uploader, so why do we have one.
2. Why bother to actually create a file instead of loading it directly into the database?

backend/flask_api.py
```diff
@@ -0,0 +1,131 @@
...
+            if (uploader.save_file_to_db(securefilename) == None):
+                return render_template('upload_success.html')
+            else:
+                return render_template('upload_error.html')
+
+
+@app.route("/uploads", methods=["GET", "POST"])
+def uploads():
+    if request.method == 'POST':
```
Drakulix: Again

backend/flask_api.py
```diff
@@ -0,0 +1,131 @@
...
+
+# receive body of elasticsearch query
+@app.route('/api/projects/search', methods=['POST'])
+def search():
+    res=elastic.es.search(index="test", doc_type="projects", body=request.json)
+    return jsonify(res)
+
+@app.route('/upload')
+def upload():
+    return render_template('upload.html')
```
Drakulix: I am usually against debug code in master, but this is very fleshed out and quite nice for testing. I actually think we should keep this for now.

backend/flask_api.py
```diff
@@ -0,0 +1,131 @@
...
+
+@app.route('/', methods=['GET'])
+def index():
+    return make_response('', 404)
+
+# receive manifest as a jsonstring
+# returns new id
+@app.route('/api/projects', methods=['POST'])
+def add_project():
+    manifest=request.json
```
Drakulix: This was the code path you were actually suppose to test and fix and this looks very similar to `save_file_to_db`.
Could you merge the two? We actually need to file upload, the frontend will already read the file into a json string.

Drakulix (changes requested):
I wish I could just merge this, looks really good, but this is not exactly, what I wanted from you two. Sorry if there was a communication issue, but @DubbleClick & @Krisselchen can you fix this in time?

henderkes:
methods are now integrated into 'api/projects', trying to upload files, get manifest from request.json or request.data, in this order

backend/flask_api.py
```diff
@@ -119,7 +118,6 @@ def get_project_by_id(project_id):
     res=coll.find_one({'_id': project_id})
     if res is None:
       return make_response('Project not found', 404)
-    return jsonify(res)
```
Drakulix: why?

.gitignore
```diff
@@ -1,4 +1,4 @@
-web/node_modules
-*.sln
+web/node_modules
```
Drakulix: whats that control character?

Drakulix (changes requested):
Soooo close

Drakulix (approved):

PR #50, opened 2017-05-27
added @krisselchen 's exception handling

henderkes:


Drakulix (changes requested):
Looks good, once an error handler is added, see:

http://flask.pocoo.org/docs/0.12/patterns/apierrors/

Drakulix (approved):

PR #51, opened 2017-05-27
Feature/disp results/iss25

Phil-Schmidt:
Searchview will now pull data from the backend and display it in a Table


Rintel (approved):

Rintel:
I've approved that PR with big concerns. While I was testing the branch I got an 404 or 500 server response but apparently it's running/working on our team members machines so it could be possible that I did something wrong. Please test this carefully and give feedback if something is strange.

PR #52, opened 2017-05-28
Backend search integrated on newest master

Drakulix:
@Rintel Please test this branch. It is based on newest master, includes the search api, but is largely untested.

@DubbleClick I need someone for review, do you have some time?


Drakulix:
Forced merge, for timely reasons ( I hate to do this )

backend/flask_api.py
```diff
@@ -128,16 +130,25 @@ def get_project_by_id(project_id):
...
+        es.delete(index="projects-index", doc_type='Project', id=project_id, refresh=True)
+    except NotFoundError:
+        if coll.delete_one({'_id':project_id}).deleted_count == 0:
+            return make_response('Project not found', 404)
+        else:
+            return make_response('Success')
     else:
-        return make_response('Project not found', 404)
+        coll.delete_one({'_id':project_id})
+        return make_response('Success')
```
henderkes, line 142: I feel like we should return a summary where the delete was successful and where not. Returning "success" if we had database inconsitency doesn't seem right.

backend/flask_api.py
```diff
@@ -128,16 +130,25 @@ def get_project_by_id(project_id):
...
 
 # receive body of elasticsearch query
 @app.route('/api/projects/search', methods=['POST'])
 def search():
-    res=elastic.es.search(index="test", doc_type="projects", body=request.json)
-    return jsonify(res)
+    try:
+        res=es.search(index="projects-index", doc_type="Project", body=request.json)
```
henderkes, line 148: I'd prefer consistency in capitalization.

backend/uploader.py
```diff
@@ -22,19 +21,24 @@ def save_file_to_db(filename):
...
+            if res['created']:
+                coll.insert_one(manifest)
+
+                print("Successfully validated file. ID is " + str(manifest['id']),file=sys.stderr)
+                print("File content is: ", file=sys.stderr)
+                print(manifest, file=sys.stderr)
+                return manifest['id']
+            else:
+                print(error,file=sys.stderr)
+                raise ApiException("ElasticSearch Index Error: \n" + str(error), 500)
```
henderkes, line 41: res['created'] also returns false if the entry was updated, Printing the error variable inside an if error=None: block is useless, especially if it hasn't necessarily been an error. I did not think about the issue of updating projects via uploader, so thank you for bringing this to my attention, I will fix it next week.

henderkes (commented):
All in all great code with no major issues. Some parts could use a rework which I will do in the next commit

PR #54, opened 2017-05-28
Filter advanced Search

PSDavidSt:
We have worked on the filtering in the advanced search. Now its possible to select filters and use them.


Rintel (approved):
This really needs some refactoring but as we are under pressure we have to merge it.
keep in mind that filtering is case sensitive right now, this is something we don't want and I will create a ticket for that.

Issue #79, opened 2017-05-31
Apply coding conventions to the existing code (31.05.2017)


Drakulix:

Leats:
Applied coding and docstring conventions to the existing python files [(in this branch)](https://github.com/Drakulix/knex/tree/hotfix/CodingConventions/iss79).
So far most of the docstrings are empty, however, because there were no comments already there. Should the original coders actually comment the code or should I read into it and add the comments in the docstrings myself?

Drakulix:
While it would have been nice, if the original authors, would have added comments, I am not sure, that will happen any time soon. It would be nice, if you could handle this and just message the authors (or me when in doubt), when you have any questions.

Drakulix:
Done by #83

PR #83, opened 2017-05-31
Application of Coding Conventions (Issue #79)

Leats:
Applied the coding conventions to the already existing python files, meaning both PEP 8 and Google docstring conventions.


Drakulix (approved):
:1st_place_medal:

PR #84, opened 2017-06-03
[WIP]  project post tests and structural idea

markwitz:
I started adding a few tests mostly post'ing simple things at the moment.
i'm just hoping to get some feedback regarding the structure


Drakulix:
I have some questions regarding how the implementation is going to work.

1. Are you actually using a python docker client to check the healthy of each container? I love that idea
2. The .travis.yml has not changed much, I assume the way this we work is by launching the docker-compose.yml and then the tests? I am assuming this, because you install the python-dependencies into travis system. Should you not create a new container for the tests instead, that is linked to the database containers? Otherwise you will not have direct access to them, which makes testing database inconsistencies much harder.

Otherwise very nice, I really like the direction this is going.

markwitz:
1. I'm trying to don't know how this will work out in the future
2. I have not configured travis properly right now:
 I am planing to start the docker container first then run the unitests directly in travis the ip and all the 
 information about docker should be gathered using the dockerpy lib i hope i cann acces the databse directly using that lib too

Drakulix:
2. This will not be possible *via* the library (thats just how docker works), but maybe you can add the required port forwardings using the library at runtime (and not change the docker-compose in the process).

Anyways, I like the structure and the ideas behind it.

tests/testmanifests/validexample1.json
```diff
@@ -41,7 +41,7 @@
     ],
     "future_work":
     [
-      "this is something related to this work",
+      "this is something related to this work",,
```
Drakulix: Woot?

Drakulix (dismissed):

Drakulix (dismissed):

Drakulix (approved):

PR #111, opened 2017-06-07
Backend/test/db testdata

Krisselchen:
Did some structural changes in json5, json5 in Arraystructure did not work out.
Testdata of evalData directory is inserted in mongodb. Function is not integrated in docker yet.


Drakulix (approved):
Looks good =)

PR #120, opened 2017-06-08
New search features and improvements

JamesComey:


backend/flask_api.py
```diff
@@ -201,5 +201,112 @@ def search():
...
+def search_simple(offset, count):
+    """Search projects
+
+    Returns:
+        res (json): JSON containing Projects and metadata
+
+    """
+    text = request.args.get("q", type=str)
+    sorting = request.args.get('sort', type=str)
+    order = request.args.get('order', type=str)
```
Drakulix, line 210: Please don't mix `@app.route` arguments and `request.args`.
Use `@app.route` for a path (like /search or /projects/<id>) and request.args for actual arguments with sensible defaults, where applicable.

So please move `offset` and `count` here as well

backend/flask_api.py
```diff
@@ -201,5 +201,112 @@ def search():
...
+    # ^3 is boosting the attribute, *_is allowing wildcards to be used
+    request_string = '{"from" : 1, "size" : 1, "query": {"multi_match" :' \
```
Drakulix: Could you create this with a real dictionary instead of loading from a string, which will happen every time at runtime and can probably not be optimized by the compiler at all and is also error prone, if you want to make adjustments later.

backend/flask_api.py
```diff
@@ -201,5 +201,112 @@ def search():
...
+        request_json["sort"][sorting] = dict()
+        request_json['sort'][sorting]['order'] = order
+    try:
+        res = es.search(index="projects-index", doc_type="Project", body=request_json)
+        return jsonify(res['hits'])
+    except RequestError as e:
+        return (str(e), 400)
+
+
+@app.route('/api/projects/search/advanced/<int:offset>/<int:count>/', methods=['GET'])
```
Drakulix: see above

backend/flask_api.py
```diff
@@ -201,5 +201,112 @@ def search():
...
+
+    Returns:
+        res (json): JSON containing Projects and metadata
+
+    """
+    text = request.args.get("q", type=str)
+    sorting = request.args.get('sort', type=str)
+    order = request.args.get('order', type=str)
+    # ^3 is boosting the attribute, *_is allowing wildcards to be used
+    request_string = '{"query": {"query_string": {"query": ""}}}'
```
Drakulix: see above

backend/flask_api.py
```diff
@@ -201,5 +201,112 @@ def search():
...
+        request_json["sort"][sorting] = dict()
+        request_json['sort'][sorting]['order'] = order
+    try:
+        res = es.search(index="projects-index", doc_type="Project", body=request_json)
+        return jsonify(res['hits'])
+    except RequestError as e:
+        return (str(e), 400)
+
+
+@app.route('/api/projects/search/tag/<int:offset>/<int:count>/', methods=['GET'])
```
Drakulix: see above

backend/flask_api.py
```diff
@@ -201,5 +201,112 @@ def search():
...
+def search_tag(offset, count):
+    """Search projects
+
+    Returns:
+        res (json): JSON containing Projects and metadata
+    """
+    tag = request.args.get('q', type=str)
+    sorting = request.args.get('sort', type=str)
+    order = request.args.get('order', type=str)
+    request_string = '{"from" : 1, "size" : 1, "query": {"bool": {"must":' \
```
Drakulix: see above

backend/flask_api.py
```diff
@@ -201,5 +201,112 @@ def search():
...
+
+@app.route('/api/projects/search/suggest/', methods=['GET'])
+def search_suggest():
+    """Suggests search improvements
+
+    Returns:
+        res (json): JSON containing suggested search terms
+    """
+    text = request.args.get('q', type=str)
+    request_string = '{"suggest": {"text" : "","phraseSuggestion" : {"phrase":' \
```
Drakulix: see above

Drakulix (changes requested):
This looks quite good, I have basically request two minor changes.

1. Please don't to arguments in `@app.route` and please don't mix them with `request.args`

2. Please use python dict and array literals instead of parsing json for static query parts. That way the syntax can be checked at compile time and more optimizations are possible.

Once that is fixed, I will happily merge it. Good job.

backend/flask_api.py
JamesComey: Yeah I am all booked out today and can do it earliest tomorrow evening.

JamesComey: Okay I pushed the changes now.

JamesComey: should be all good now

Drakulix (dismissed):
You could have used the dictionary literal, but this is fine as well. I will merge sometime today, once I rebased this on current master.

Drakulix:
@JamesComey I could not help myself, but clean that up. Could please verify that this is still doing what it should. Also this needs some tests, then I could have merged that myself. Please make this your next task.

backend/flask_api.py
```diff
@@ -199,5 +199,174 @@ def search():
...
+    sorting = request.args.get('sort', type=str)
+    order = request.args.get('order', type=str)
+    offset = request.args.get('offset', type=int)
+    if offset is None:
+        offset = 0
+    count = request.args.get('count', type=int)
+    if count is None:
+        count = 10
+
+    # ^3 is boosting the attribute, *_is allowing wildcards to be used
```
henderkes, line 218: I think this should be ^2

henderkes:
search_simple and search_advanced seem to be 95% the same, could look into summing that up
looks nice otherwise, it's too late to test right now but I will tomorrow if needed

Drakulix (approved):
@JamesComey confirmed function

PR #121, opened 2017-06-09
users collection setup and insert_user()

aTepliak:
collection for user login and registration data, and prototype insert_user()


backend/flask_api.py
```diff
@@ -20,6 +21,24 @@
 client = MongoClient('mongodb:27017')
 db = client.knexDB
 coll = db.projects
+users = db.users
+
+def insert_user(firstname, lastname, email, password_hash,bio="",bookmarks=""):
```
Drakulix: Could you refactor this in an endpoint call? PUT /api/users?

Drakulix (changes requested):
Good start. Please remove the pycache files. :)

backend/flask_api.py
```diff
@@ -37,7 +59,7 @@
 def index():
     """Index of knex
     """
-    return make_response('', 404)
+    return  make_response('', 404)
```
henderkes, line 62: please fix

backend/flask_api.py
```diff
@@ -20,6 +21,27 @@
...
+
+
+#function prototype for  insert user
+#should the password be hashed here?
+def insert_user(firstname, lastname, email, password_hash,bio="",bookmarks=""):
+    try:
+        users.insert_one(
+        {
+        "email": email,
+        "password_hash": password_hash, #bcrypt.hashpw(password_hash.encode('utf-8'), bcrypt.gensalt()),
```
henderkes, line 34: this line is too long to pass pep-8

henderkes:
seeing as we want to use flask-security for the user and login system I do not think we can use the database to handle users like this. see flask-mongoengine and flask-security mongoengine introductions as example

PR #123, opened 2017-06-09
Added several tests for backend for post, get and delete

Leats:
### Questions/problems I have:
Added __test_misformed_content_type__ in the class _TestPOST_, which tests if a valid json with a wrong header (in this case I used image/gif, which definitely should not work) still raises an error. So far, this test fails.

Added __test_empty_post__, which tests for 400 when one posts an empty body. However, that one does return a 500 with the error „error: :1:1 expecting '0X‘“.

### Other tests run through fine for me:
Added __test_missformed_json_error__, which tests posting invalid json and mismatching json for Bad Request and Validation error.

Added __test_unkown_id__, which test if there’s a Bad Request error when one wants to delete a project by an invalid ID (so far the ID is just ‚invalid‘, no checking if the ID is truly not used yet. Is that necessary as the format of the ID is wrong?). 

Added __test_success__, which tests if a project can get deleted properly. For that I’ll first upload a project (same as the test_success_json), and then delete that exact project by the ID, to make sure the test can run properly even if the project database was empty before. 

Added __test_unknown_id__, which tests for 404 when one tries to get a project with an invalid ID. 

Added __test_success_getid__, which tests if GET functions properly (after posting a new project). Tests for both 200 and the same title and analysis to make sure the right project gets returned.


tests/api/test_projects.py
```diff
@@ -53,36 +58,134 @@ def test_validtion_error(self, flask_api_url, pytestconfig):
...
-        assert True
 
 
 class TestDELETE(object):
+
     def test_unknown_id(self, flask_api_url):
-        assert True
+        """ Test for 404 when a project with unknown ID is to be deleted.
+        """
+        invalid_id = "invalid"
```
Drakulix: It would be nice to a valid format here. You can just generate a new v4 uuid, a collision is (neither) impossible.

The more I think about it, this seems like to separate tests. Having "unknown" and "invalid" id checks would be nice, thanks for pointing out that issue.

tests/api/test_projects.py
```diff
@@ -53,36 +58,134 @@ def test_validtion_error(self, flask_api_url, pytestconfig):
...
         assert True
 
 
 class TestGET(object):
+
     def test_unknown_id(self, flask_api_url):
-        assert True
+        """ Test for 404 when a project with unknown ID is to be deleted.
+        """
+        invalid_id = "invalid"
```
Drakulix: See above

Drakulix (changes requested):
Looks very good, thanks. The two issue you have raised are bugs in the implementation and a good example, why we want tests.

For both of these separate issues should be filled, thanks.

Leats:
Updated tests. I made a __test_unknown_id__ and a __test_invalid_id__ for both get and delete.
Unknown generates a random uuid4.
Invalid uses an id in an invalid format, as in the first version.

The two issues for the found bugs I will create now.

Thank you for reviewing!

Leats:
All checks pass now with the two hotfixes!

Drakulix (approved):

PR #124, opened 2017-06-09
Backend/bug/wrong uuid/iss122

Krisselchen:
The dumps()-method rewrote the UUID into bad style. Fixed this with a for-loop. The UUIDs are returned correct now.

### DIFF
--- a/backend/flask_api.py
+++ b/backend/flask_api.py
@@ -3,7 +3,6 @@ import os
 import sys
 
 import json5
-from bson.json_util import dumps
 from elasticsearch import Elasticsearch
 from elasticsearch.exceptions import RequestError
 from flask import Flask, request, jsonify, make_response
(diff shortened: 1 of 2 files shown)


backend/flask_api.py
```diff
@@ -143,7 +142,12 @@ def get_projects():
     else:
         return make_response('Invalid parameters', 400)
 
-    res = make_response(dumps(res))
+    entries = res[:]
+    resArr = []
+    for entry in entries:
+        resArr.append(entry)
```
henderkes, line 148: sum those lines up in
resArr = [x for x in res[:]] please

edit: FE needs this today iiuc so i'll approve and change it in my pr for later

henderkes (changes requested):

henderkes (approved):

backend/flask_api.py
Drakulix: For future cases, where a quick merge is desired and the code itself is solid (like this one), I strongly recommend adding a `TODO` comment to these lines really quick or fill an issue and link it here, it is okay to change other peoples branches for that reason. Tracking those smaller required changes to keep the code clean(er) is otherwise really difficult.

Issue #112, opened 2017-06-07
[BE] Mark existing functions for login_required and required roles of users in Flask-Security


Drakulix:
Project push & get should require a login
Project put (#104) should only work for user matching author
Project delete admin only for now

henderkes:
#126 now made project delete admin only, put only for admin or user matching autho

post (push?) and get not changed yet, to not fail current tests

Issue #110, opened 2017-06-07
[BE] Add default admin user with default password


Drakulix:
Add default admin user
This should be a one-time job for database initialization (seperate script?)

henderkes:
initial setup on first Flask app run now contains creating a default user "user@knex.com" and a default admin "admin@knex.com"

Issue #109, opened 2017-06-07
[BE] Setup user and admin roles in Flask-Security


Drakulix:
We have two roles (users and admin) those need to be setup in flask-security

PR #125, opened 2017-06-10
[WIP] detailed exception handling for the api/projects POST method

henderkes:
I'm not sure if returning a tuple of (list(ids), list(ApiException)) is best design but couldn't come up with a better idea quickly.
What exactly did you have in mind @markwitz ?
This references Issue #82

PR #126, opened 2017-06-10
[WIP] user and admin roles, initial user creation, permission example

henderkes:
References issues #112 #110 #109


Drakulix (commented):
Very good start. I am not sure however if we should merge this in the current state, when there is no way to login, the initial user creation is this simple and this seems to depend on code of another PR of yours. Also there are pyc-files.

But really solid, I like the way this is going!

henderkes:
yeah, I accidentally created this branch from within the project-post-adjustment branch which resulted in the first 8 commits, it's however not reliant on the other PR, I can do a rebase from master if needed

henderkes:
rewrote history, should be conform with master now and not use the changes of #125

henderkes:
for some reason it still lets me access @login_required protected pages without logging in.
I will look into that and build a basic login page as soon as I find time

also references issue #130 now. collaborators are welcome, feel free to assign yourself

backend/flask_api.py
```diff
@@ -30,6 +29,74 @@
 ALLOWED_EXTENSIONS = {'txt', 'json', 'json5'}
 app.config['UPLOAD_FOLDER'] = ''
 app.config['MAX_CONTENT_PATH'] = 1000000  # 100.000 byte = 100kb
+app.config['MONGODB_DB'] = 'knexDB'
+app.config['MONGODB_HOST'] = 'mongodb'
+app.config['MONGODB_PORT'] = 27017
+
+es = Elasticsearch([{'host': 'elasticsearch', 'port': 9200}])
+
+client = MongoClient('mongodb:27017')
```
markwitz, line 38: `client = MongoClient('mongodb:27017')` should'nt the client get the connection by using the config?
```
+app.config['MONGODB_DB'] = 'knexDB'
+app.config['MONGODB_HOST'] = 'mongodb'
+app.config['MONGODB_PORT'] = 27017
```
if u are startig to use the config at this point we should use it as well

henderkes: had to do that for the MongoEngine, but yes I agree, a good idea to use it for the project db client as well

would look like an abomination though, I guess there's no direct constructor

Issue #127, opened 2017-06-10
[BE] Content-Type Header doesn't get checked (POSTing to /api/projects)


Leats:
When POSTing to _/api/projects_ with Content-Type Header != "application/json" or "application/json5", __but__ a valid JSON it should return a 400 Bad Request because of the wrong header.

However, it does return a 200. 

(Referencing test from issue #97 and written test, which is currently in pull request #123 )

henderkes:
How can this produce a 200 status code if that is nowhere in the code?
It currently stores valid jsons anyway no matter the content header.

Drakulix:
200 is default, if you return something successfully @DubbleClick

henderkes:
oh, in that case our error handling was simply shit, rewrote this quickly to raise an exception instead of returning a response

Leats:
@DubbleClick Was this one fixed as well? I thought only #128 was fixed, or did you close the wrong issue? Because in the master branch the test_misformed_content_type test still fails. test_empty_post is passing.

Drakulix:
Reopen until this is verified as fixed.

henderkes:
Actually, we cannot return a 400 error if the content header is not application/json(5) as the file upload will kick in then, I will check for no files and wrong content header then

henderkes:
should be fixed, please test @Leats

Leats:
Fixed, thank you!

PR #140, opened 2017-06-11
hotfix for #127

henderkes:
self explanatory

### DIFF
--- a/backend/flask_api.py
+++ b/backend/flask_api.py
@@ -55,19 +55,17 @@ def add_project():
                     newId = uploader.save_file_to_db(securefilename)
                     # represent original filename
                     successful_files.append(file.filename + " " + str(newId))
-                except Exception as e:
+                except ApiException as e:
                     unsuccessful_files.append(file.filename + str(e))
 
                 print("Successful files: ", successful_files, '\n', file=sys.stderr)
                 print("Unsuccessful files: ", unsuccessful_files, '\n', file=sys.stderr)
-        return """<!doctype html>
-                    <title>Upload multiple files</title>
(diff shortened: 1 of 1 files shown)


Leats (approved):
All tests pass! Thank you!

backend/flask_api.py
```diff
@@ -55,19 +55,17 @@ def add_project():
...
-                    <body>Successful files: """ + ', '.join(e for e in successful_files) + '<br />' + """
-                    Unsuccessful files: """ + ', '.join(e for e in unsuccessful_files) + """
-                    </body>"""
+        return make_response('Successful files: ' +
+                             ', '.join(e for e in successful_files) +
+                             '<br />' + " Unsuccessful files: " +
+                             ', '.join(e for e in unsuccessful_files))
 
-    else:  # no files attached
+    elif request.content_type == 'application/json' or request.content_type == 'application/json5':
```
Drakulix, line 68: Oh no, this is ugly. This mean 'application.json' can pass, if the contents are actually 'application/json5' (vise versa is not a big deal and allowed).

Drakulix (commented):
Why was that merged?

backend/flask_api.py
henderkes: with 'application/json' set `if request.json` will be the case, the `elif request.data:` will not be executed. The problem you described can therefore not happen. json5 content in request.json will result in an ApiException being raised

henderkes:
I restored the branch to make it prettier down the road, I actually don't like the design this way myself (double if check). this also references issue #141 

@Drakulix you are welcome to revert the merge if you are too unhappy with the current code, it should however not present a functionality issue

PR #142, opened 2017-06-11
[Backend][WIP]created structure moved route function to blueprints

### LINKED ISSUE #78: Refactor overall backend structure
- [x] Move `app.route` functions into `projects.py` in a new `api folder` (see structure below)
- [x] Check that it is still working and fix if needed

- [x] Rename the flask_api.py into __init__.py (see structure below)

- [x] Make sure everything works again

The overall structure of the project should look like this:
- backend
    - __init__.py
    - api
        - projects.py
        - (later users.py, etc)

markwitz:
I want to resolve #78:
moved overall structure to:
...
- backend
 - api
 - init.py
 - Dockerfile
 - requirements.txt
   - exceptions
     - apiexceptions.py
   - projects_helper
      - uploader.py
   - projects.py 
    - [ ] add usermanagement here
  ...
I moved the projects route to a blueprint and initialised the database connection for each request i'm not sure if that's the right way to to do it or if it should be opened on flask initialisation?
the database connection for es and mongo are now strored in the flask.g variable to be made accessible.

I'm not to happy with the current state of uploader because right now it handles uploading files and simply inputting the json data from the request so naming is inconvenient here. Because it's mostly handling database storage.

I have not testet the file upload but the test of the current state of master all passed on my machine and on travis.


Drakulix:
I don't think recreating the database connection on every request is the way to go and very performant. I am kinda leaning towards keeping them global.

henderkes:
How about the equivalent to a c++ namespace to get rid of globals? Edit: They are, @markwitz  did you mean @before_**first_**request by any chance?

markwitz:
from my understanding flask.g is only stored for every request made, so i moved the connection to a global but still still made the connection available using flask.g which makes it easier accessing them from other blueprints.

markwitz:
@DubbleClick  Database connection is now created on first request it is somehow named before_first_app_request even thou the blueprint ist called differently. I'm not sure if the connection should maybe stored in the flask.config? hard figuring out best practice here.

backend/api/projects.py
```diff
@@ -22,16 +22,25 @@ def init_global_manifest_validator():
...
+
+@projects.before_app_first_request
+def init_gloabl_mongoclienth():
+    global mongoclient
+    mongoclient = MongoClient('mongodb:27017')
 
 @projects.before_request
 def init_gloabl_elasticsearch():
-    g.es = Elasticsearch([{'host': 'elasticsearch', 'port': 9200}])
+    g.es = es
```
henderkes: I am not sure why you keep the assignments then if they are globally available anyway?

Drakulix: Same

Drakulix:
Superseeded by #218

Issue #133, opened 2017-06-10
[BE] Test and potentially fix non utf-8 encoded request bodies


Drakulix:
Currently passing non utf-8 encoded request bodies should result in a 500 Internal Server Error, because they are most likely raising an exception.

This should be checked by tests, because this should instead result in a 400 (Bad Request) and be fixed, if this is actually an issue.

For functions hit by this problem search for `.decode('utf-8')` calls.

henderkes:
I will have to change the content-type checks for this, with a specific encoding that should usually be specified in there, too.

henderkes:
request.data is generally unformatted (type: str, not unicode or bytes) in python3
therefore the first and only way to check it is to see if decode('utf-8') throws an exception and catch the correct one.

Drakulix:
Yes, but currently this will result in an internal server error. I just want that exception to be catched and then returned as an 400 status code

henderkes:
It was previously caught as a general exception, fixed now.

PR #146, opened 2017-06-12
check utf-8 encoding, fix for content-type

henderkes:
header content-type can now contain more than only 'application/json[5]', such as 'application/json; charset: utf-8' see #133

### DIFF
--- a/backend/flask_api.py
+++ b/backend/flask_api.py
@@ -84,8 +84,9 @@ def add_project():
 
         except ApiException as e:
             raise e
-        except UnicodeError as ue:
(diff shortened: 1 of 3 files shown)


Drakulix:
Maybe reword the error message to 'Only utf-8 compatible charsets are supported, the request body does not appear to be utf-8.'. The message might otherwise be quite confusing, if you e.g. specified charset=ascii in the content-type and the returned error is 'no valid utf-8'.

henderkes:
Ugh, this took longer than expected but it should be ready and working now

Drakulix (approved):

PR #147, opened 2017-06-12
Backend/feature/update projects/iss104

Krisselchen:
The update function is working now for json and json5 manifests. Automated tests are not implemented yet. They hopefully will be done tomorrow . FE ist waiting for that feature. Therefore created that PR anyway.

### DIFF
--- a/backend/flask_api.py
+++ b/backend/flask_api.py
@@ -233,7 +233,8 @@ def update_project(project_id):
     except ApiException as error:
         raise error
     except UnicodeDecodeError as unicodeerr:
-        raise ApiException("utf8 decoding not possible", 400, unicodeerr)
+        raise ApiException('Only utf-8 compatible charsets are supported, ' +
+                           'the request body does not appear to be utf-8.', 400)
     except Exception as err:
         raise ApiException(str(err), 500)
 
--- a/backend/flask_api.py
+++ b/backend/flask_api.py
@@ -4,6 +4,7 @@ Defines API points and starts the application
 
 import os
 import sys
+import time
 import json
 import json5
 
@@ -12,7 +13,7 @@ from elasticsearch.exceptions import RequestError
 from flask import Flask, request, jsonify, make_response
 from flask_cors import CORS
 from jsonschema import FormatChecker, Draft4Validator
-from pymongo import MongoClient
+from pymongo import MongoClient, ReturnDocument
 from werkzeug.utils import secure_filename
 
 import uploader
@@ -180,14 +181,14 @@ def delete_project(project_id):
             return make_response('Success')
 
 
-#TODO
+# TODO
 @app.route('/api/projects/<uuid:project_id>', methods=['PUT'])
 def update_project(project_id):
     try:
         res = coll.find_one({'_id': project_id})
         if res is None:
             raise ApiException("Project not found", 404)
-        #TODO checkmimetype
+        # TODO checkmimetype
         elif request.is_json:
             manifest = request.json
             if manifest['_id'] != str(project_id):
@@ -195,15 +196,16 @@ def update_project(project_id):
             is_valid = validator.is_valid(manifest)
             print(is_valid, file=sys.stderr)
             if is_valid:
+                print(manifest['_id'])
                 manifest['_id'] = project_id
-                entry['date_update'] = time.strftime("%Y-%m-%d")
+                manifest['date_update'] = time.strftime("%Y-%m-%d")
                 coll.find_one_and_replace({'_id': project_id}, manifest,
                                           return_document=ReturnDocument.AFTER)
                 print("mongo replaced: ", file=sys.stderr)
                 es.index(index="projects-index", doc_type='Project',
-                          id=entry["_id"], refresh=True, body={})
+                          id=manifest["_id"], refresh=True, body={})
                 print("Successfully replaced content: ", file=sys.stderr)
-                print(entry, file=sys.stderr)
+                print(manifest, file=sys.stderr)
                 return make_response('Success')
             else:
                 v = validator.iter_errors(manifest)
--- a/backend/flask_api.py
+++ b/backend/flask_api.py
@@ -198,7 +198,7 @@ def update_project(project_id):
         res = coll.find_one({'_id': project_id})
         if res is None:
             raise ApiException("Project not found", 404)
-        elif request.is_json or request.content_type == "application/json5;charset: utf-8":
+        elif request.is_json or "application/json5" in request.content_type:
             if request.is_json:
                 manifest = request.get_json()
                 if manifest['_id'] != str(project_id):
@@ -221,9 +221,10 @@ def update_project(project_id):
                 print("Successfully replaced content: ", file=sys.stderr)
                 print(manifest, file=sys.stderr)
                 return make_response('Success')
-            elif validator.iter_errors(manifest) is not None:
-                validation_errors = [error for error in sorted(validator.iter_errors(manifest))]
-                raise ApiException("Validation Error: \n" + str(is_valid), 400, validation_errors)
+            else:
+                validation_errs = [error for error in sorted(validator.iter_errors(manifest))]
+                if validation_errs is not None:
+                    raise ApiException("Validation Error: \n" + str(is_valid), 400, validation_errs)
         else:
             raise ApiException("Manifest had wrong format", 400)
     except ApiException as e:
--- a/backend/flask_api.py
+++ b/backend/flask_api.py
@@ -215,20 +215,25 @@ def update_project(project_id):
                 manifest['date_last_updated'] = time.strftime("%Y-%m-%d")
                 coll.find_one_and_replace({'_id': project_id}, manifest,
                                           return_document=ReturnDocument.AFTER)
-                print("mongo replaced: ", file=sys.stderr)
-                es.index(index="projects-index", doc_type='Project',
-                         id=manifest["_id"], refresh=True, body={})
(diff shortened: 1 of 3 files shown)


backend/flask_api.py
```diff
@@ -191,6 +192,62 @@ def delete_project(project_id):
...
+        response: Success response
+                  or 404 if project is not found
+                  or 409 if project_id differs from manifestID
+                  or 500 in case of validation error
+    """
+    try:
+        res = coll.find_one({'_id': project_id})
+        if res is None:
+            raise ApiException("Project not found", 404)
+        elif request.is_json or request.headers['content-type'] == "application/json5":
```
henderkes: please check 'application/json5' in request.content_type instead, as it might be "application/json5;charset: utf-8"

backend/flask_api.py
```diff
@@ -191,6 +192,62 @@ def delete_project(project_id):
...
+                es.index(index="projects-index", doc_type='Project',
+                         id=manifest["_id"], refresh=True, body={})
+                print("Successfully replaced content: ", file=sys.stderr)
+                print(manifest, file=sys.stderr)
+                return make_response('Success')
+            else:
+                v = validator.iter_errors(manifest)
+                if v is not None:
+                    validation_error = []
+                    for error in sorted(validator.iter_errors(manifest), key=str):
```
henderkes: no need to call validator.iter_errors(manifest) a second time, also v as variable name should be avoided as it's in an iterator over errors

backend/flask_api.py
```diff
@@ -191,6 +192,62 @@ def delete_project(project_id):
...
+                  or 409 if project_id differs from manifestID
+                  or 500 in case of validation error
+    """
+    try:
+        res = coll.find_one({'_id': project_id})
+        if res is None:
+            raise ApiException("Project not found", 404)
+        elif request.is_json or request.headers['content-type'] == "application/json5":
+            if request.is_json:
+                manifest = request.json
```
henderkes: use request.get_json() instead (and set silent), also see the last comment

backend/flask_api.py
```diff
@@ -191,6 +192,62 @@ def delete_project(project_id):
...
+                    for error in sorted(validator.iter_errors(manifest), key=str):
+                        print(error.message, file=sys.stderr)
+                        validation_error.append(error.message)
+                raise ApiException("Validation Error: \n" + str(is_valid), 400, validation_error)
+        else:
+            raise ApiException("Manifest had wrong format", 400)
+    except ApiException as e:
+        raise e
+    except Exception as err:
+        raise ApiException(str(err), 500)
```
henderkes: get_json() or str.decode('utf-8') might both raise exceptions (UnicodeDecodeError), please check for that before catching general exception and return status code 400 then (wrong request encoding)

backend/flask_api.py
```diff
@@ -191,6 +192,62 @@ def delete_project(project_id):
...
+            raise ApiException("Project not found", 404)
+        elif request.is_json or request.headers['content-type'] == "application/json5":
+            if request.is_json:
+                manifest = request.json
+                if manifest['_id'] != str(project_id):
+                    raise ApiException("Updated project owns different id", 409)
+            else:
+                manifest = json5.loads(request.data.decode("utf-8"))
+                if '_id' in manifest:
+                    if manifest['_id'] != str(project_id):
```
henderkes, line 209: shouldn't it also raise an exception if '_id' is not a field in manifest?

henderkes (changes requested):
a few suggestions and error handling, looks good otherwise

backend/flask_api.py
Krisselchen: I did not do that because of the possibillity of users updating their local json/json5, they had before and give that one as input. Then they do not have the db id, but it could be a possible use case. It would also be  an update an the id has to be set.

Maybe we could check for equivalent title to be sure everything is right in that case....

backend/flask_api.py
```diff
@@ -191,6 +192,62 @@ def delete_project(project_id):
...
+                         id=manifest["_id"], refresh=True, body={})
+                print("Successfully replaced content: ", file=sys.stderr)
+                print(manifest, file=sys.stderr)
+                return make_response('Success')
+            else:
+                v = validator.iter_errors(manifest)
+                if v is not None:
+                    validation_error = []
+                    for error in sorted(validator.iter_errors(manifest), key=str):
+                        print(error.message, file=sys.stderr)
```
henderkes: without this print it's possible to just
`validation_errors = [error for error in sorted(validator.iter_errors(manifest))]`

backend/flask_api.py
Krisselchen: suggestions instead of v?

henderkes: see below, "errors" would be an appropriate name

backend/flask_api.py
henderkes: I understand, but I'm not sure if we should allow an update call then as we don't have other unique identifiers as is.
@Drakulix should know what to do about that

backend/flask_api.py
```diff
@@ -209,9 +209,9 @@ def update_project(project_id):
         res = coll.find_one({'_id': project_id})
         if res is None:
             raise ApiException("Project not found", 404)
-        elif request.is_json or request.headers['content-type'] == "application/json5":
+        elif request.is_json or request.content_type == "application/json5;charset: utf-8":
```
henderkes: you have to check for:
`elif request.is_json or 'application/json5' in request.content_type:`

specifying the charset or other things in the header content is optional, but if you check for == (equals) you will not catch any variance there

backend/flask_api.py
```diff
@@ -232,14 +232,9 @@ def update_project(project_id):
...
-                v = validator.iter_errors(manifest)
-                if v is not None:
-                    validation_error = []
-                    for error in sorted(validator.iter_errors(manifest), key=str):
-                        print(error.message, file=sys.stderr)
-                        validation_error.append(error.message)
-                raise ApiException("Validation Error: \n" + str(is_valid), 400, validation_error)
+            elif validator.iter_errors(manifest) is not None:
+                validation_errors = [error for error in sorted(validator.iter_errors(manifest))]
+                raise ApiException("Validation Error: \n" + str(is_valid), 400, validation_errors)
```
henderkes: this still calls validator.iter_errors(manifest) twice, what I meant is:

```
validation_errors = [error for error in sorted(validator.iter_errors(manifest))]
if validation_errors:
    raise ApiException("Validation Error: \n" + str(is_valid), 400, validation_errors)
```
I don't know how costly the call is, suspect it to take quite a bit though with long manifests

backend/flask_api.py
Drakulix: I actually don't see a problem with this code, this is exactly what I expected.

Drakulix (dismissed):
Good job. Showing warnings or similar is the frontends responsibility as there is no good way to do this via HTTP. Changing the title must be possible, this is the reason, why we have another unique identifier. So a "stupid" replace call is exactly what I wanted.

backend/flask_api.py
```diff
@@ -191,6 +192,58 @@ def delete_project(project_id):
...
+                return make_response('Success')
+            else:
+                validation_errs = [error for error in sorted(validator.iter_errors(manifest))]
+                if validation_errs is not None:
+                    raise ApiException("Validation Error: \n" + str(is_valid), 400, validation_errs)
+        else:
+            raise ApiException("Manifest had wrong format", 400)
+    except ApiException as e:
+        raise e
+    except Exception as err:
```
henderkes: please check for UnicodeDecodeError and then we're good!

henderkes (commented):
looks good to me then, travis passed too so I'll approve when UnicodeDecodeError exception is handled!

henderkes:
please be careful to not override the latest hotfix changes when rebasing this with master later

backend/flask_api.py
```diff
@@ -191,6 +192,63 @@ def delete_project(project_id):
...
+            else:
+                validation_errs = [error for error in sorted(validator.iter_errors(manifest))]
+                if validation_errs is not None:
+                    raise ApiException("Validation Error: \n" + str(is_valid), 400, validation_errs)
+        else:
+            raise ApiException("Manifest had wrong format", 400)
+    except ApiException as error:
+        raise error
+    except UnicodeDecodeError as unicodeerr:
+        raise ApiException("utf8 decoding not possible", 400, unicodeerr)
```
henderkes: utf-8 :(
@Drakulix asked to have the unicode error say:
```
raise ApiException('Only utf-8 compatible charsets are supported, ' +
                               'the request body does not appear to be utf-8.', 400)
```

backend/flask_api.py
```diff
@@ -191,6 +192,63 @@ def delete_project(project_id):
...
+            is_valid = validator.is_valid(manifest)
+            if is_valid:
+                print("manifest validated", file=sys.stderr)
+                manifest['_id'] = project_id
+                manifest['date_last_updated'] = time.strftime("%Y-%m-%d")
+                coll.find_one_and_replace({'_id': project_id}, manifest,
+                                          return_document=ReturnDocument.AFTER)
+                print("mongo replaced:", file=sys.stderr)
+                print(manifest, file=sys.stderr)
+                manifest.pop('_id', None)
```
henderkes, line 220: why not store the document to ES first?

henderkes (commented):
minor changes, otherwise good!

henderkes (dismissed):

Drakulix (approved):

PR #148, opened 2017-06-13
Backend/user api endpoints/iss113

pdler:
Everything should work ...


backend/.idea/backend.iml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="PYTHON_MODULE" version="4">
+  <component name="NewModuleRootManager">
+    <content url="file://$MODULE_DIR$" />
+    <orderEntry type="jdk" jdkName="Python 3.5.3 (/usr/bin/python3.5)" jdkType="Python SDK" />
```
henderkes: hardcoding the python version to 3.5.3 does not seem to be a good idea to me

henderkes (changes requested):

backend/flask_api.py
```diff
@@ -11,19 +12,31 @@
...
+    UserMixin, RoleMixin, login_required, roles_required, login_user, logout_user
+from passlib.hash import pbkdf2_sha256
 
 es = Elasticsearch([{'host': 'elasticsearch', 'port': 9200}])
 
 client = MongoClient('mongodb:27017')
 db = client.knexDB
 coll = db.projects
-
+coll_user = db.user
```
henderkes: wait, why?

backend/flask_api.py
```diff
@@ -40,6 +53,90 @@ def index():
...
+
+# Setup Flask-Security
+user_datastore = MongoEngineUserDatastore(db, User, Role)
+security = Security(app, user_datastore)
+app.url_map.converters['email'] = EmailConverter
+
+@app.before_first_request
+def initialize_users():
+    user_role = user_datastore.find_or_create_role('user')
+    user_datastore.create_user(email='user@knex.com', password=pbkdf2_sha256.hash("user"), roles=[user_role])
```
henderkes: this is how users are created, what is the db.users collection for?

henderkes (changes requested):

backend/.idea/backend.iml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Drakulix: please add `.idea` to your .gitignore

backend/flask_api.py
```diff
@@ -11,19 +12,31 @@
...
 from werkzeug.utils import secure_filename
-
+from werkzeug.routing import BaseConverter
+from mongoengine.fields import UUIDField
 import uploader
 from apiexception import ApiException
+from flask_mongoengine import MongoEngine
+from flask_security import Security, MongoEngineUserDatastore, \
+    UserMixin, RoleMixin, login_required, roles_required, login_user, logout_user
+from passlib.hash import pbkdf2_sha256
```
Drakulix: See #163

backend/flask_api.py
```diff
@@ -201,5 +298,99 @@ def search():
...
+        raise e
+    except Exception as err:
+        raise ApiException(str(err), 500)
+
+
+
+
+
+@app.route('/api/users/', methods=['PUT'])
+@roles_required('admin')
```
Drakulix: Allow a user to update itself

Drakulix (changes requested):
Additional tasks:

- [ ] Allow only users to replace their project now that #147 is merged
- [ ] Make Project upload @login_required
- [ ] Make project delete @admin required
- [ ] Comment out all @admin/login_required annotations for merge, so we have time to fixup travis, write tests and for the frontend to catch up until we enable this requirement.

backend/flask_api.py
Drakulix: Also this should be `/api/users/<email>` to be really RESTful

backend/flask_api.py
```diff
@@ -201,5 +298,99 @@ def search():
...
+    try:
+        user = request.get_json()
+
+        # still without json validation
+        # a new user does not have bookmarks
+        hash = pbkdf2_sha256.hash(user["password"])
+        role = user_datastore.find_or_create_role(user['role'])
+        res = coll_user.find({'email' : user["email"]})
+        if res is not None:
+            return make_response('User already exists',500)
```
Drakulix: Please make this a `400` for `Bad Request` instead of a 500.

henderkes:
Please take a look at drakulix' requested changes and also pay attention to pep8 (in /knex/ directory execute "pip install pep8" and then "pep8 ."

The "manual" user collection still exists, we wanted to get rid of that

henderkes:
with the default users removed we'll need an initialisation script to create at least one admin user

backend/flask_api.py
```diff
@@ -118,20 +118,20 @@ def login():
     password = request.form['password']
     user = user_datastore.get_user(email)
     if user is None:
-        return 'Username oder Password invalid'
+        return make_response('Username oder Password invalid', '500')
```
henderkes: please don't pass the status code as string but as integer

henderkes (changes requested):
please remove __pycache__ stuff, merge current master into this branch and fix pep8 coding conventions, it will be easier to have a detailed look then

pdler:
ready to merge..

backend/flask_api.py
```diff
@@ -43,6 +50,85 @@ def index():
...
+# Setup Flask-Security
+user_datastore = MongoEngineUserDatastore(db, User, Role)
+security = Security(app, user_datastore)
+app.url_map.converters['email'] = EmailConverter
+
+
+@app.before_first_request
+def initialize_users():
+    user_role = user_datastore.find_or_create_role('user')
+    pw = encrypt_password("user")
```
Drakulix: `create_user` already encrypts the password, this call should only be needed for the update call, where we directly set the password.

backend/flask_api.py
```diff
@@ -43,6 +50,85 @@ def index():
...
+    user_datastore.\
+        create_user(email='user@knex.com', password=pw, roles=[user_role])
+    admin_role = user_datastore.find_or_create_role('admin')
+    pw = encrypt_password("admin")
```
Drakulix: See above

backend/flask_api.py
```diff
@@ -416,11 +517,77 @@ def search_suggest():
...
+        # still without json validation
+        # a new user does not have bookmarks
+        role = user_datastore.find_or_create_role(user['role'])
+        # if res is not None:
+        # return make_response('User already exists',500)
+
+        user_datastore.create_user(first_name=user["first name"],
+                                   last_name=user["last name"],
+                                   email=user["email"],
+                                   password=encrypt_password(user["password"]),
```
Drakulix, line 533: See above

Drakulix (changes requested):
Almost perfect

backend/flask_api.py
pdler: No it does not seem so...
If I remove the encrypt_password function, the password is stored as plaintext....(SECURITY_PASSWORD_HASH and SECURITY_PASSWORD_SALT are set...)

Drakulix (dismissed):

backend/flask_api.py
Drakulix: Seems you are right, this is a little weird, but okay.

backend/flask_api.py
```diff
@@ -161,7 +161,8 @@ def add_project():
                 return_ids = uploader.save_manifest_to_db(
                     json5.loads(request.data.decode('utf-8')))
             else:
-                raise ApiException("Wrong content header and no files attached", 400)
+                raise ApiException("Wrong content header\
+                 and no files attached", 400)
```
henderkes: pep8 discourages line breaks in statements and encourages (same line) operators instead

backend/flask_api.py
```diff
@@ -86,10 +85,11 @@ class User(db.Document, UserMixin):
 
 
 class EmailConverter(BaseConverter):
-    regex = r"([a-z0-9!#$%&'*+\/=?^_`{|}~-]\
-    +(?:\.[a-z0-9!#$%&'*+\/=?^_`""{|}~-]+)\
-    *(@|\sat\s)(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])\
-    ?(\.|""\sdot\s))+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?)"
+    regex = r"([a-z0-9!#$%&'*+\/=?^_`{|}~-]+(?:\."+\
+            "[a-z0-9!#$%&'*+\/=?^_`""{|}~-]+)"+\
+            "*(@|\sat\s)(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?"+\
```
henderkes: I think there is no reason for the backslash here, operators should come with a single whitespace

backend/flask_api.py
```diff
@@ -8,7 +8,6 @@
 import time
 import json
 import json5
-
```
henderkes: ??? please keep that in, the encouraged form is:

-absolute imports- (import library)

-relative imports- (from library import X)

-own absolute imports- (import projectfile)
-own relative imports- (from projectfile import X)

backend/flask_api.py
```diff
@@ -86,10 +85,11 @@ class User(db.Document, UserMixin):
...
 class EmailConverter(BaseConverter):
-    regex = r"([a-z0-9!#$%&'*+\/=?^_`{|}~-]\
-    +(?:\.[a-z0-9!#$%&'*+\/=?^_`""{|}~-]+)\
-    *(@|\sat\s)(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])\
-    ?(\.|""\sdot\s))+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?)"
+    regex = r"([a-z0-9!#$%&'*+\/=?^_`{|}~-]+(?:\."+\
+            "[a-z0-9!#$%&'*+\/=?^_`""{|}~-]+)"+\
+            "*(@|\sat\s)(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?"+\
+            "(\.|""\sdot\s))+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?)"
+
```
henderkes: this is a line break too much, pep8 again

henderkes:
The code looks good now, but I don't think we should merge this with so many "discouraged styling issues". You also import json5 twice, line breaks are missing or too much, Backslash operator is used very often though discouraged. I will fix these myself now but we should pay more attention in the future, given that we promised a consequent coding style.

DubbleClick:
The code looks good now, but I don't think we should merge this with so many "discouraged styling issues". You also import json5 twice, line breaks are missing or too much, Backslash operator is used very often though discouraged. I will fix these myself now but we should pay more attention in the future, given that we promised a consequent coding style.

henderkes (dismissed):

henderkes (approved):
added comments

PR #149, opened 2017-06-13
filled body in ES

Krisselchen:
added body


henderkes:
It's correct now, we're both trying to look into what lets the tests fail now. It's strange as the return didn't change...

henderkes:
closed for now until we found a fix

PR #161, opened 2017-06-14
[Backend] mongo connector working now 

### LINKED ISSUE #87: [BE] [Impl] Connecting MongoDB and Elasticsearch
### Research : Syncing MongoDB and Elasticsearch
I've looked into tools for syncing, and these were the mentioned options:
 
#### [mongo-connector](https://pypi.python.org/pypi/mongo-connector) 
creates a pipeline from a MongoDB cluster to one or more target systems, like Elasticsearch. It synchronizes data in MongoDB to the target then tails the MongoDB oplog, keeping up with operations in MongoDB in real-time. Can be easily installed with pip and seems to be the popular option since mongodb river is deprecated (for river read further). 
See also the wiki of mongo-connector: [Usage with Elasticsearch](https://github.com/mongodb-labs/mongo-connector/wiki/Usage-with-Elasticsearch)
 
#### [Transporter](https://github.com/compose/transporter) 
is for synchronizing data e.g. from mongoDB to elasticsearch, however, that does come as a one time export and would need to be modified to keep it in sync (feature only in beta).
 
#### [logstash](https://www.elastic.co/guide/en/logstash/current/introduction.html)
Logstash is an open source data collection engine with real-time pipelining capabilities. Logstash can dynamically unify data from disparate sources and normalize the data into destinations of your choice.
Has a plugin for output to elasticsearch, but only a community-maintained non-verified plugin for mongodb input. Probably possible but doesn’t seem like the easiest option.
 
[MongoDB river](https://github.com/richardwilly98/elasticsearch-river-mongodb) seems to have been a popular choice, as well. But that one is deprecated and won’t work with the newest MongoDB version.
 
### Conclusion
In the end I’d personally stick to mongo-connector, which had been suggested at first, too. Easily set up and actually made exactly for our purpose.

markwitz:
resolve #87 
I had to create a separate container starting the replica set dunno if there is a better way to avoid this but it's working like this for me. Es is getting updated when I insert to mongo.
Still needs testing:
- [x] the id issue how is it handled  by the connector ?
- [x] can we find projects by ID?
- [x]  at what rate are we syncing ? and can we configure it?
- [x] what about including excluding collections? account data?


backend/mongo_connector/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.4
```
henderkes: better remove the subversion specification

henderkes (commented):
the whole thing with the replica set being required is normal, I will have a deeper look at this on the weekend but it looks very good for now.

could anyone working on linux/docker-toolbox try to pull the branch, set it up locally and post back results?

backend/mongo_connector/Dockerfile
markwitz: its stable for 3.4 so i specified it

markwitz:
I don't really know whats happening but the test that has passed at first will no longer run on travis even those without changes. I cant recreate locally

https://travis-ci.org/Drakulix/knex/jobs/242851069 passed but when re reran for the same commit it would not pass and throw me some error

henderkes:
Apparently "all" tests fail so there's most likely some internal server error going on (with the connector right after storing a manifest in mongodb?). We simply need better debugging tools than pytest, flask prints and travis output.

markwitz:
the container that should initialise the repSet is constantly restarting but i cant really tell why because i can't reproduce the error on my machine

henderkes:
Strange enough this works wonderfully for me locally, too

henderkes:
image

hmm..

henderkes:
Okay, so mongodb service fails and cannot be connected to if the --replSet option is specified. Therefore the replicaset image won't run and the connector won't work. The question is why the --replSet option makes mongodb fail in travis but not locally. Please someone with linux test this,

henderkes:
We can set the '--auto-commit-interval=0' flag to make every write operation in mongodb to be instantly synced to elasticsearch.

The mongodb['_id'] field becomes the elasticsearch id field.
excluding collections is done in the connector config in the backend/mongo_connector folder. We should probably exclude the user stuff, but I'm not sure in which collection it is?

henderkes:
Excluded the user collection, everything works flawlessly now except for travis...

henderkes:
From Mongoconnector docs: 

Elasticsearch Indexes, Mappings, and Types

> Mongo Connector gives each MongoDB database its own index in Elasticsearch. Each MongoDB collection becomes its own mapping type. For example, documents from the collection kittens in the database animals will put into the animals index in Elasticsearch with a mapping type of kittens

the index knexdb is not generated, therefore the connector is not working yet.

backend/mongo_connector/mongo-connector-config.json
```diff
@@ -12,7 +12,8 @@
     "docManagers" : [
       {
         "docManager" : "elastic2_doc_manager",
-        "targetURL" : "elasticsearch:9200"
+        "targetURL" : "elasticsearch:9200",
+	"autoCommitInterval" : 0
```
henderkes: missing a space

Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM ubuntu:latest
```
henderkes, line 1: we should really replace this by something more lightweight in the future, but it does the job for now

backend/flask_api.py
```diff
@@ -31,34 +31,33 @@
...
 es = Elasticsearch([{'host': 'elasticsearch', 'port': 9200}])
 client = MongoClient('mongodb:27017')
-db = client.knexDB
+db = client.knexdb
 coll = db.projects
-app = Flask(__name__)
-CORS(app)
 
-with app.open_resource("manifest_schema.json") as schema_file:
+with open("manifest_schema.json") as schema_file:
```
henderkes: we should revert this to app.open_resource() and see what the problem with str/bytes was

backend/flask_api.py
```diff
@@ -31,34 +31,33 @@
...
     schema = json.load(schema_file)
 validator = Draft4Validator(schema, format_checker=FormatChecker())
 
-ALLOWED_EXTENSIONS = {'txt', 'json', 'json5'}
-app.config['UPLOAD_FOLDER'] = ''
-app.config['MAX_CONTENT_PATH'] = 1000000  # 100.000 byte = 100kb
-
```
henderkes: One blank line too much, but really not important

backend/flask_api.py
```diff
@@ -31,34 +31,33 @@
...
-login_manager = LoginManager()
-login_manager.init_app(app)
+app = Flask(__name__, static_url_path='')
+CORS(app)
 
 # MongoDB Config
-app.config['MONGODB_DB'] = 'knexDB'
+app.config['DEBUG'] = True
+app.config['SECRET_KEY'] = 'super-secret'
```
henderkes: should probably replace this is something more professional

backend/flask_api.py
```diff
@@ -31,34 +31,33 @@
...
+app.config['MONGODB_DB'] = 'knexdb'
 app.config['MONGODB_HOST'] = 'mongodb'
 app.config['MONGODB_PORT'] = 27017
 app.config['SECURITY_PASSWORD_HASH'] = 'pbkdf2_sha512'
 app.config['SECURITY_PASSWORD_SALT'] = 'THISISMYOWNSALT'
```
henderkes, line 43: see above

backend/flask_api.py
```diff
@@ -335,8 +329,7 @@ def search():
         res (json): Body of the Query
     """
     try:
-        res = es.search(index="projects-index",
-                        doc_type="Project", body=request.json)
+        res = es.search(index="knexdb", doc_type="Project", body=request.json)
```
henderkes: we should remove doc_type specification and this should be request.get_json()

backend/mongodb_replica_setup/setup.sh
```diff
@@ -0,0 +1,21 @@
...
+echo SETUP.sh time now: `date +"%T" `
+mongo --host mongodb:27017 <<EOF
+   var cfg = {
+        "_id": "projectsRepSet",
+
+        "members": [
+            {
+                "_id": 0,
+                "host": "mongodb:27017",
+		"priority": 1,
```
henderkes: missing one space here?

docker-compose.yml
```diff
@@ -3,17 +3,38 @@ version: '3'
...
+        restart: "on-failure"
+
+    mongodb_replica_setup:
+        build: ./backend/mongodb_replica_setup
+        depends_on:
+            - "mongodb"
 
     elasticsearch:
-        image: elasticsearch
+        image: docker.elastic.co/elasticsearch/elasticsearch:5.4.1
```
henderkes, line 17: reason for this change?

docker-compose.yml
```diff
@@ -3,17 +3,38 @@ version: '3'
...
+        environment:
+            - "http.host=0.0.0.0"
+            - "transport.host=127.0.0.1"
+            - "xpack.security.enabled=false"
+            - "action.auto_create_index=true"
         ports:
             - "9200:9200"
+        restart: "on-failure"
```
henderkes, line 25: we should test whether an elasticsearch error inducing a shutdown returns a failure code, mongo-connector for example doesn't and has to be on "always"

tests/api/test_database_connection.py
```diff
@@ -18,3 +20,23 @@ def test_empty_database(flask_api_url):
...
+def test_consistency(mongo_client, elastic_client):
+    project_id = uuid.uuid4()
+    dummy_project = {"_id": project_id, "name": "dummy_file"}
+    mongo_client.projects.insert_one(dummy_project)
+    mongo_result = mongo_client.projects.find_one(project_id)
+    print("mongo:", mongo_result)
+    # Test if document is in mongo
+    assert mongo_result == dummy_project
+    # wait for the refreshed elastic database
+    time.sleep(30)
```
henderkes: sleep(30) is a little long, 5 seconds should do

tests/api/test_projects.py
```diff
@@ -220,16 +252,26 @@ def test_success(self, flask_api_url, pytestconfig):
         print(post_response.text)
         for id in post_response.json():
             assert UUID(id, version=4)
+        project_id = post_response.json()[0]
```
henderkes: this line is unnecessary, we should instead assert post_response.status_code == 200

henderkes (commented):
overall very well, I will adress most changes myself, but would like input on the docker-compose config one

backend/flask_api.py
```diff
@@ -22,43 +22,40 @@
...
 from apiexception import ApiException
 
 
-es = Elasticsearch([{'host': 'elasticsearch', 'port': 9200}])
-client = MongoClient('mongodb:27017')
-db = client.knexDB
-coll = db.projects
-app = Flask(__name__)
+# Create app
+app = Flask(__name__, static_url_path='')
```
markwitz, line 33: what does static_url_path do?

backend/flask_api.py
```diff
@@ -22,43 +22,40 @@
...
-
-
-# Create app
-app = Flask(__name__)
+# MongoDB Config
 app.config['DEBUG'] = True
 app.config['SECRET_KEY'] = 'super-secret'
-
-login_manager = LoginManager()
-login_manager.init_app(app)
```
markwitz, line 54: why are we dropping this?

backend/flask_api.py
```diff
@@ -82,7 +79,7 @@ class User(db.Document, UserMixin):
 
 class EmailConverter(BaseConverter):
     regex = r"([a-z0-9!#$%&'*+\/=?^_`{|}~-]+(?:\." +\
-            r"[a-z0-9!#$%&'*+\/=?^_`""{|}~-]+)" +\
+            r"[a-z0-9!#$%&'*+\/=?^_`"r"{|}~-]+)" +\
```
markwitz, line 82: why are we adding the "r" here?

markwitz (commented):
I had only a few questions

backend/flask_api.py
henderkes: make it a raw string literal, it isn't necessary there but goes in style with the other r's I added (which I had to, \ and other letters have to be escaped otherwise which they weren't)

backend/flask_api.py
henderkes: no idea, this isn't from me, from docs:

> static_url_path – can be used to specify a different path for the static files on the web. Defaults to the name of the static_folder folder.

> static_folder – the folder with static files that should be served at static_url_path. Defaults to the 'static' folder in the root path of the application.

backend/flask_api.py
henderkes: it's not dropped, it's relocated

markwitz: ok where is it now?

henderkes (approved):
GREEN

backend/flask_api.py
Drakulix: If of interested, this basically means, that static files should be served from the root of the webside, instead of "/static/" as is default.

PR #167, opened 2017-06-14
[Frontend] Connected the project discovery page to the current api

Phil-Schmidt:
The project discovery view will now properly display projects from the backend using the new api functions.


PSDavidSt (approved):
testet and it worked

PR #168, opened 2017-06-14
Updated Font of Logo

Leats:
I did try to find a cleaner font which will scale in a nicer way.
Please tell me if it's okay like that of if a different font would be better/ I should change anything else.


henderkes:
looks much cleaner

Rintel (approved):

PR #174, opened 2017-06-16
Frontend/advanced search/adjusting to api

Rintel:


Phil-Schmidt:
I am encountering a bug where sometimes multiple "GET /api/projects/search/simple/?q=music&offset=0&count=4" are fired AFTER the request containing the filter is send.
This causes the filtered result to be overwritten and not rendered.
My guess is that this might be a concurrency issue, because the console outputs a lot of warnings that imply that you are changing the state during the render-function.
Also it usually happens when I am typing a lot of characters quickly into an empty textarea of the filter which causes slowdown on my old ass machine but never, when I am editing a preexisting filter by a single character.

PSDavidSt (approved):

PR #180, opened 2017-06-19
[Frontend] Create Project by Json

Phil-Schmidt:
build basic functionality for importing external JSON files in the project creation form.

Upon pressing the Enterkey in the URL-field of the project-creation page, you will be redirected to the project creation form. All of the form inputs will be filled with the data from the json.
No form of error checking or handling has been implemented.  

At this point I have still encountered some cross-origin related errors with some specific urls.
Files from Github should be fine though.
If you encounter any issues, try downloading and enabling the chrome-plugin Allow-Control-Allow-Origin: *
to override these for the time being.


web/src/components/views/CreateProjectByPattern.jsx
```diff
@@ -113,22 +122,205 @@ export default class UploadByPattern extends React.Component {
...
+          type: "string",
+          title: "Status",
+          default: this.state.status
+        },
+        url: {
+          type: "string",
+          title: "Github URL",
+          format: "uri"
+        },
+        url_two: {
```
Rintel: This has been changed because of how the validation of the backend works. Therefore we had to remove the second URL. 

Right now it looks like this:
`    url: {
      type: "array",
      title: "URL",
      items: {
        type: "string"
      }
    },`

Rintel:
As no one else feels the need to review this branch I'm gonna approve it anyway

Rintel (dismissed):

web/src/components/views/CreateProjectByPattern.jsx
```diff
@@ -261,15 +267,12 @@ export default class UploadByPattern extends React.Component {
...
-        url_two: {
-          type: "string",
-          title: "Other URL",
-          format: "uri",
-          default: this.state.url
+          type: "array",
+          title: "URL",
+          items: {
+            type: "string"
+          },
```
Drakulix, line 274: This will need to be extended soon, as we promised the alternative of adding descriptions to the links. Which means this can also be an object with the keys "title" and "url". Maybe you wanna fix that now?

Anyway this is not critical for merge.

web/src/components/views/CreateProjectByPattern.jsx
```diff
@@ -282,14 +285,14 @@ export default class UploadByPattern extends React.Component {
...
-  
+
   render() {
     if(this.state.anySet){
       return(
         <div className="container">
           <div className="innerContainer">
             <div className="headerCreation">Create New Project</div>
-              <Form schema={this.dlschema()}
+              <Form schema={this.dlschema()}yvb
```
Drakulix: yvb?

Drakulix (changes requested):
Thanks for doing this.
I think you got some weird typo in there "yvb" (see comment). Otherwise nice.

web/src/components/views/CreateProjectByPattern.jsx
Rintel: I'm gonna tackle this issue by the end of the week. There are too many more important issues under construction right now.

Drakulix (approved):

PR #184, opened 2017-06-20
Test security/iss165

aTepliak:
tests for flask security functions


.idea/codeStyleSettings.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
henderkes: remove this file please and then add *.idea to the .gitignore

tests/api/test_security.py
```diff
@@ -0,0 +1,100 @@
+import requests
+import unittest
+import os
+
+
+class TestSecurity(object):
+    def setUp(self):
```
henderkes: no camelCase allowed, please change this to setup(self) or set_up(self). Better yet would be `__init__(self)` though
https://docs.python.org/2/reference/datamodel.html#object.__init__

tests/api/test_security.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def test_main_page(self, flask_api_url):
+        response = requests.get(flask_api_url + '/')
+        assert response.status_code == 200
+
+    def test_login_fake_user(self, flask_api_url):
+        response = requests.post(flask_api_url + '/api/users/login',
+                                 data=dict(email='user1', password='password'))
+        print(response)
+        assert response.reason == 'INTERNAL SERVER ERROR'
```
henderkes: This is very poor behaviour (not the test methods fault, but this has to be changed too), a login attempt with wrong user/password should not lead to an internal server error but a 400 code

tests/api/test_security.py
```diff
@@ -0,0 +1,100 @@
...
+        response = requests.post(flask_api_url + '/api/users/login',
+                                 data=dict(email='admin@knex.com', password="admin"))
+        assert response.reason == 'OK'
+        assert response.status_code == 200
+
+    def test_login_real_user_wrong_psswd(self, flask_api_url):
+        response = requests.post(flask_api_url + '/api/users/login',
+                                 data=dict(email='admin', password='a'))
+        assert response.reason == 'INTERNAL SERVER ERROR'
+        assert response.status_code == 500
```
henderkes: see above

tests/api/test_security.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def test_logout(self, flask_api_url):
+        response = requests.get(flask_api_url + '/api/users/login',
+                                data=dict(email='user@knex.com', password="user"))
+        response = requests.get(flask_api_url + '/api/users/logout')
+        assert response.reason == 'OK'
+        assert response.status_code == 200
+
+    def test_access_login_requiered_logged(self, pytestconfig, flask_api_url):
```
henderkes: remove the "e" please, it's "required"

tests/api/test_security.py
```diff
@@ -0,0 +1,100 @@
...
+            str(pytestconfig.rootdir),
+            'tests',
+            'testmanifests',
+            'validexample0.json5'
+        )
+        with open(test_manifest, 'r', encoding='utf-8') as tf:
+            data = str(tf.read().replace('\n', ''))
+        response = requests.post(flask_api_url + "/api/projects",
+                                 data=data.encode('utf-8'),
+                                 headers={'Content-Type': 'application/json5'})
```
henderkes: nice :)

tests/api/test_security.py
```diff
@@ -0,0 +1,100 @@
...
+                                data=dict(email='user@knex.com'))
+        assert response.status_code == 404  # 200?
+
+    def test_update_user_not_exists(self, flask_api_url):
+        response = requests.get(flask_api_url + '/api/users/',
+                                data=dict(email='unknownuser@knex.com'))
+        assert response.status_code == 404
+
+
+'''
```
henderkes: We currently use """ for multiline comments, "some string" for strings and 'variable' for directives and such

henderkes (changes requested):

Drakulix:
@DubbleClick is right, otherwise this looks good! Thanks for tackling this @aTepliak.

aTepliak:
for logint/ fake login test, the "BAD REQUEST" with   400 code, should be returned in respective functions in flask_api.py.
Now there is  "Internal server error" with 500 code

Drakulix:
@aTepliak Your .idea files are still in there, but otherwise this looks good now. However we should fix the user endpoints before merging this.

tests/api/test_security.py
```diff
@@ -16,9 +16,9 @@ def test_main_page(self, flask_api_url):
     def test_login_fake_user(self, flask_api_url):
         response = requests.post(flask_api_url + '/api/users/login',
                                  data=dict(email='user1', password='password'))
-        print(response)
+
         assert response.reason == 'INTERNAL SERVER ERROR'
```
henderkes: This first assert has to removed or changed, too

tests/api/test_security.py
```diff
@@ -30,7 +30,7 @@ def test_login_real_user_wrong_psswd(self, flask_api_url):
         response = requests.post(flask_api_url + '/api/users/login',
                                  data=dict(email='admin', password='a'))
         assert response.reason == 'INTERNAL SERVER ERROR'
```
henderkes: This also shouldn't be an internal server error

henderkes (commented):
couple more things, rest looks good!

tests/api/test_security.py
aTepliak: oki doki

aTepliak:
@DubbleClick  I removed the .idea files and travis fails in a very strange manner now

henderkes:
That's because the branch is not up to date with master, docker-engine no longer exists and therefore travis can't install it. Do you want me to rebase?

Edit: Master was merged into this, a fault I did too often myself.
@Drakulix do you think it's worth attempting a rebase?

henderkes:
To revive this a bit if this should ever land in master I made some changes and cleaned up a little. The /api/users endpoints were a little broken too and I applied some fixes here. @aTepliak this still needs testing and further work, please continue with this as I have a few other issues to solve.

backend/api/users.py
```diff
@@ -144,11 +141,8 @@ def delete_bookmarks(id):
...
 @login_required
 def get_bookmarks():
-    user = current_user
-    if user is None:
-        return make_response("No current user detected ", 400)
-    res = g.user_datastore.get_user(user.email)
-    if res is None:
+    res = g.user_datastore.get_user(current_user.email.to_python())
+    if not res:
         return make_response("Unknown User with Email-address: " +
```
henderkes, line 146: I'm not even sure if this makes sense. It's the current_user anyway and this is @login_required so there is literally no chance that the user is unknown. If for any reason the user_datastore can't find the user (?) this should raise an internal server error / apiexception.

Other input is welcome, just my thought.

backend/api/users.py
```diff
@@ -102,19 +102,19 @@ def get_user(mail):
...
 
 @users.route('/api/users/bookmarks/<uuid:id>', methods=['POST'])
 @login_required
-def insert_bookmarks(id):
-    user = current_user
-    res = g.user_datastore.get_user(user['email'])
-    if res is None:
-        return make_response("Unknown User with Email-address: ", 400)
+def insert_bookmark(id):
+    res = g.user_datastore.get_user(current_user.email.to_python())
```
henderkes, line 114: is this even necessary? Can't we just use res = current_user? I don't see why not, get_user and current_user are both of type User(Document, UserMixin)

henderkes (changes requested):
In general, should we allow admins to bookmark projects for other users? I guess not, but just to make sure.

PR #186, opened 2017-06-21
[BE] Tests get all projects 

markwitz:
resolve #186 
added a test that tests if get all returns manifests and added some database cleanup


tests/api/conftest.py
```diff
@@ -32,3 +38,42 @@ def mongo_client():
...
+    with open(path_to_schema, mode='r') as schema_file:
+        schema = json.load(schema_file)
+    validator = Draft4Validator(schema, format_checker=FormatChecker())
+    return validator
+
+
+@pytest.yield_fixture(autouse=True)
+def run_around_tests(mongo_client):
+    # Code that will run before your test, for example:
+    # files_before = # ... do something to check the existing files
```
Drakulix: This part of the comments seems copy&pasted and completely unrelated

tests/api/test_database_connection.py
```diff
@@ -13,8 +14,10 @@ def test_no_elastic(flask_api_url, docker_client):
     assert True
 
 
-def test_empty_database(flask_api_url):
+def test_empty_database(flask_api_url, mongo_client):
     """Sample test."""
+    result = mongo_client.projects.delete_many({})
+    print("Database cleaned: ", result)
     response = requests.get(flask_api_url + "/api/projects")
     assert response.text == "There are no projects"
```
Drakulix: Although this test is correct, I am not certain this is the best way handle this case. I prefer the response to be an empty json array in that cases and the error code to be 200, could you quickly change that?

tests/api/test_projects.py
```diff
@@ -291,8 +288,22 @@ def test_invalid_id(self, flask_api_url):
...
         assert response.status_code == 404
 
-    def test_success_getall(self, flask_api_url):
-        assert True
+    def test_success_getall(self, flask_api_url, manifest_validator,
+                            mongo_client, enter_data_using_post):
+        response = requests.get(flask_api_url + "/api/projects")
+        print(response.status_code)
+        assert response.status_code == 200
+        projects = response.json()
```
Drakulix: with your new yield_fixture will there be any contents in the db here?

Drakulix (changes requested):
Overall good stuff, just some questions and nitpicks.

tests/api/test_projects.py
markwitz: it should cleanup before the test and after the test

Drakulix: So then there are no projects in here, right? Whats the purpose of this test then.

Drakulix (dismissed):

markwitz:
fixing response coder for getprojects on empty database

markwitz:
@Drakulix could you rereview?

Drakulix (dismissed):
This looks nice now.

Drakulix:
@markwitz I apologize for the mess on the master currently, but could you rebase? I will then merge this asap. Looks fine now.

henderkes:
rebased this as it was only my changes this was conflicting with

henderkes (approved):

PR #218, opened 2017-06-21
Refactor flask_api.py to use Blueprints

Drakulix:
This touches almost everything, but needs to be put into master asap so work can continue. Please review soon, but carefully and with caution.


backend/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+def init_global_elasticsearch():
+    global ES
+    ES = Elasticsearch([{'host': 'elasticsearch', 'port': 9200}])
+
+@app.before_request
+def set_global_elasticsearch():
+    g.es = ES
+
+@app.before_first_request
+def init_gloabl_mongoclient():
```
henderkes: typo, please fix

backend/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+    global LOGINMANAGER
+    LOGINMANAGER = LoginManager()
+    LOGINMANAGER.init_app(app)
+
+@app.before_request
+def init_global_manifest_validator():
+    with app.open_resource("manifest_schema.json", mode='r') as schema_file:
+        schema = json.load(schema_file)
+        global VALIDATOR
+        VALIDATOR = Draft4Validator(schema, format_checker=FormatChecker())
```
henderkes, line 72: this doesn't have to be called every request since it's a global, not member of flasks "g" object

backend/api/projects.py
```diff
@@ -0,0 +1,187 @@
...
+projects = Blueprint('api_projects', __name__)
+
+# @login_required
+@projects.route('/api/projects', methods=['POST'])
+def add_project():
+    """Receive manifest as a jsonstring and return new ID
+    """
+    successful_ids = []
+    unsuccessful_files = []
+    uploaded_files = request.files.getlist("file[]")
```
henderkes: ehh, not on topic but before I forget it this should probably be 'file[]' to fit the rest of the code

backend/api/projects.py
```diff
@@ -0,0 +1,187 @@
...
+    """Updates Project by ID
+
+    Args:
+        project_id, updated manifest in json/json5 format
+
+    Returns:
+        response: Success response
+                  or 404 if project is not found
+                  or 409 if project_id differs from manifestID
+                  or 500 in case of validation error
```
henderkes: this should not return 500 (internal server error) in case of a validation error @Krisselchen can you fix this after this is in master?

backend/api/users.py
```diff
@@ -0,0 +1,93 @@
...
+        # a new user does not have bookmarks
+        role = g.user_datastore.find_or_create_role(user['role'])
+        # if res is not None:
+        # return make_response('User already exists',500)
+
+        g.user_datastore.create_user(first_name=user["first name"],
+                                   last_name=user["last name"],
+                                   email=user["email"],
+                                   password=encrypt_password(user["password"]),
+                                   bio=user["bio"], roles=[role])
```
henderkes: under-indented

backend/api/users.py
```diff
@@ -0,0 +1,93 @@
...
+
+    except ApiException as e:
+        raise e
+    except Exception as err:
+        raise ApiException(str(err), 500)
+
+
+@users.route('/api/users/', methods=['PUT'])
+# @roles_required('admin')
+def updateUser():
```
henderkes: this should be update_user

backend/api/users.py
```diff
@@ -0,0 +1,93 @@
...
+    res.save()
+    res = make_response(dumps(res))
+    res.headers['Content-Type'] = 'application/json'
+
+    return make_response("User with email: " + user['email'] + " updated", 200)
+
+
+@users.route('/api/users/<email:mail>', methods=['GET'])
+@login_required
+def getUser(mail):
```
henderkes: and get_user

backend/api/users.py
```diff
@@ -0,0 +1,93 @@
...
+
+@users.route('/api/users/logout')
+def logout():
+    logout_user()
+    return make_response('Logged out', 200)
+
+
+@users.route('/api/users', methods=['PUT'])
+# @roles_required('admin')
+def createUser():
```
henderkes: create_user

henderkes (changes requested):
travis doesn't run yet, but I didn't check what's wrong yet. pep8 will fail either way at the moment.

I like the project structure, +1

Drakulix:
@DubbleClick should all be done now, pls check again.

backend/__init__.py
markwitz: it says before first request

markwitz:
when did we change travis setup to download like that `sudo apt-get -y -o Dpkg::Options::="--force-confnew` ? we need to fix that so we can be sure everything works fine

markwitz (commented):
i really like the structure and the approach to init int the __init__ file I will approve once we fix travis

.travis.yml
```diff
@@ -1,5 +1,7 @@
 sudo: required
 
+group: edge
```
henderkes, line 3: I wanted to suggest that 😠

backend/__init__.py
```diff
@@ -127,12 +141,15 @@ def handle_invalid_usage(error):
     return response
 
 # Flask URL methods
+
+
```
henderkes: two new lines after seem to be a bit much after a comment, especially when there should be two new lines after the end of a method (instead of one like here)

backend/api/projects.py
```diff
@@ -20,13 +20,15 @@
 projects = Blueprint('api_projects', __name__)
 
 # @login_required
+
+
```
henderkes: I think we should remove the new lines at least here, the method should later be @login_required without new lines, it's not a section start as before.
It should actually be moved below the @projects.route() statement

henderkes (commented):
few more changes, then good to go

backend/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+app.register_blueprint(users)
+app.register_blueprint(search)
+
+
+@app.route('/<path:dummy>', methods=['GET'])
+def index():
+    """Index of knex
+    """
+    return app.send_static_file('index.html')
+
```
henderkes, line 161: relocate that above the app.register_blueprints please

backend/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+    Returns:
+        response (json): Error in json format
+    """
+    response = jsonify(error.to_dict())
+    response.status_code = error.status_code
+    return response
+
+# Flask URL methods
+
+
```
henderkes: same as before

henderkes (changes requested):

backend/api/projects.py
```diff
@@ -0,0 +1,189 @@
...
+from api.helper import uploader
+from api.helper.apiexception import ApiException
+
+
+projects = Blueprint('api_projects', __name__)
+
+# @login_required
+
+
+@projects.route('/api/projects', methods=['POST'])
```
henderkes: still the same as well

backend/__init__.py
Drakulix: Must be under the blueprint to not catch the api endpoints.

backend/api/projects.py
```diff
@@ -0,0 +1,188 @@
...
+
+from api.helper import uploader
+from api.helper.apiexception import ApiException
+
+
+projects = Blueprint('api_projects', __name__)
+
+
+# @login_required
+@projects.route('/api/projects', methods=['POST'])
```
henderkes: swap the two lines, then it's good to go

henderkes (dismissed):

henderkes (approved):

PR #220, opened 2017-06-21
Put setup of date_creation before verifying step

Krisselchen:
I did changes in uploader.py. The "date_creation" is a required field. Setting of this field after verifying, would make no sense. In case of missing field it has to be set before verifying otherwise verification is failing anyway and it will never be set. I am sorry, but I could not test. Docker did not build the master. I have to check, what is the problem. Maybe you could check if these changes are running.


Drakulix (dismissed):

henderkes:
Oh yeah, I wanted to rework the entire method last weekend already, thank you for quickly fixing this.

henderkes (dismissed):

Drakulix (approved):

PR #228, opened 2017-06-24
Fe/feature/usermanagement

JamesComey:
Usermanagement works but will be improved later on. Please merge asap, we need do have usermanagement and I can't fix anything until sunday evening but other people(s life) may depend on this... :)

### DIFF
--- a/backend/__init__.py
+++ b/backend/__init__.py
@@ -33,6 +33,9 @@ app.config['MAX_CONTENT_PATH'] = 1000000  # 100.000 byte = 100kb
 global DB
 DB = MongoEngine(app)
 
+LOGINMANAGER = LoginManager()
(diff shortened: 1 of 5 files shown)


Drakulix (commented):
This looks quite good, but I give @Rintel the decision to merge here, because this has one big flaw, that will need to be fixed at some point.

As far as I understand the code, this tracks the login status locally on the client side with another cookie. There is currently no way to detect, that flask may has logged you out, the application might still assume to be logged in.

This is even very likely, as the expire-time is quite high and we don't pass the remember flag to flask.

This is fine as a first draft, but we can absolutely not deploy this, like that and that would need to be fixed asap.

Drakulix (approved):

PR #243, opened 2017-06-28
uncomment @login_required 

henderkes:
and require admin role, change create_user() to POST.

create_user() and update_user() were both put which makes no sense, changed create_user to POST to follow the general scheme

commit runs fine locally

### DIFF
--- a/backend/api/users.py
+++ b/backend/api/users.py
@@ -18,7 +18,7 @@ def login():
 
     if verify_password(password, user["password"]):
         login_user(user)
-        return make_response('Login successful', 200)
+        return make_response("Login successful", 200)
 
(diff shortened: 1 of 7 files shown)


Drakulix:
Please update the tests as well.

backend/api/users.py
```diff
@@ -57,8 +56,8 @@ def create_user():
         raise ApiException(str(err), 500)
 
 
-@users.route('/api/users', methods=['PUT'])
-# @roles_required('admin')
+@users.route('/api/users/', methods=['PUT'])
```
henderkes: I'm not sure about this

Drakulix (approved):

PR #244, opened 2017-06-28
Fix direct url links

Drakulix:
`static_url_path` disallows our catch-all index route to work, causing the app only to be initialized correctly when started with the index path `/`. Accessing path directly like `/discovery` or project links without loading the index at first did fail.
This PR fixes this by handling the index in the 404 error catcher. This is a little bit hacky, but still the best solution I could find.


henderkes:
I really don't like this, it looks so hacky. We'd also have to adjust it if we ever implement calls not starting with /api/. Can't we fix the issue by first starting it with '/' in the backend and then handling the first request?
Also this won't fix the issue if the first call is something like '/api/index', would it?

Eh...

Drakulix:
> Can't we fix the issue by first starting it with '/' in the backend and then handling the first request?

I don't understand what you mean by that and I expect the answer to be "No."
We need to serve `index.html` for every path, that is not a static file and is not handled by our api calls. Yes this is hacky, but I don't think there is another way.

> Also this won't fix the issue if the first call is something like '/api/index', would it?

That would return a `404` like I would expect it to.
What this is supposed to fix is calling `/discovery` without calling the index first (you can easily reproduce this on current master). Without it this will return a 404 from the backend without the frontend being displayed. This is much more critical for project paths, which will also not work, but sharing a project can only be done via link.

The alternative is to whitelist all frontend routes, instead of blacklisting everything that starts with `api`. Would you like that more? That would also be quite straight-forward, but means more work for the frontend, whenever they want to add a new route.
Also like this the frontend can display a custom 404 error page, when a non-"/api" url is called but not found, which will not work this whitelisting.

henderkes (commented):
I suppose you are right, if the frontend is on board with this solution I have no objections.

Drakulix:
Alright, then this is mostly looking for anyone of the frontend team to test this and confirm this is working as intended.

henderkes:
@Rintel does the merge mean this is ready to be merged? from BE side this can be approved

Rintel (approved):

Rintel:
@DubbleClick yep, I don't have any remarks.

PR #245, opened 2017-06-28
[FE] Refactor side bar component, remove unnecessary pages

Drakulix:
- [x] Fleshed out side bar component
- [x] Removed now unnecessary pages
- [x] Adjusted index.js so it can load and display views directly.
- [x] Quickly checked that everything appears to be working
- [x] Went to sleep

Please review and report back.


web/src/components/common/SideBar.jsx
```diff
@@ -1,61 +1,46 @@
...
         </div>
-      </div>
     );
   }
 }
 
-export default SearchBar;
+const SideBarEntry = ({to, name, active}) => {
+    return (
+        <li className={"list-group-item" + (active ? "active" : "")}>
```
mharasic, line 39: A whitespace in the condition before "active" is missing. 
If active is evaluated to true, the resulting class will be "list-groupitemactive"

Rintel (approved):
👍

web/src/components/common/SideBar.jsx
```diff
@@ -1,61 +1,46 @@
...
         createProject: 'Create New Project',
+        bookmarks: 'Bookmarks',
+        profile: 'Profile',
         adminArea: 'Admin Area'
       }
     };
   }
 
-  render() {
+  render({ location }) {
```
Rintel: You're passing a string but it's expecting an object here. So no curly brackets.

PR #246, opened 2017-06-29
[BE]archive projects

markwitz:
- [x] get all projects without archived 
- [x] get one project without archived 
- [x] get all archived projects 
- [x] get one archived project
- [x] unarchive project 
- [ ] filter ES search so it only returnes archved or not archived results
- [x] Tested?
- i implemented archiving as an optional field in the manifest 
- default is to not display archived projects
- archived projects will be displayed by using a different route 
- unarchiving can right now be done by using the update call

### DIFF
--- a/backend/api/projects.py
+++ b/backend/api/projects.py
@@ -224,8 +224,8 @@ def update_project(project_id):
                                                 return_document=ReturnDocument.AFTER)
                 g.notify_users(
                     list(set(
-                        [author.email for author in manifest['authors']
-                            if author.email != current_user['email']] +
+                        [author['email'] for author in manifest['authors']
+                            if author['email'] != current_user['email']] +
                         g.users_with_bookmark(str(manifest['_id']))
                     )), "Project was updated", manifest['title'],
                     '/projects/' + str(manifest['_id']))
--- a/backend/api/projects.py
+++ b/backend/api/projects.py
@@ -2,7 +2,6 @@
 Defines API points and starts the application
 """
 
-import sys
 import time
 import uuid
 import json
@@ -92,9 +91,8 @@ def get_projects():
     limit = request.args.get('limit', type=int)
     skip = request.args.get('skip', type=int)
     archived = request.args.get('archived', type=str, default='false')
-    print(request.query_string, file=sys.stderr)
-
-    print(archived, file=sys.stderr)
+    if archived not in ['true', 'false', 'mixed']:
+        return make_response('Invalid parameters', 400)
     argc = len(request.args)
 
     if g.projects.count() == 0:
@@ -104,18 +102,14 @@ def get_projects():
         query = {'archived': True}
     elif archived == 'false':
         query = {'archived': False}
-    elif archived == 'mixed':
-        query = {}
     if limit and skip:
         res = g.projects.find(query, limit=limit, skip=skip)
     elif limit:
         res = g.projects.find(query, limit=limit)
     elif skip:
         res = g.projects.find(query, skip=skip)
-    elif archived:
-        res = g.projects.find(query)
     else:
-        return make_response('Invalid parameters', 400)
+        res = g.projects.find(query)
 
     try:
         res = [x for x in res[:]]
@@ -160,7 +154,7 @@ def get_project_by_id(project_id):
     Returns:
         res (json): Project corresponding to the ID
     """
-    archived = request.args.get('archived', type=str, default="false")
+    archived = request.args.get('archived', type=str, default="mixed")
     if archived == 'true':
         res = g.projects.find_one({'_id': project_id, 'archived': True})
     elif archived == 'false':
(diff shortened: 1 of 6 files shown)


Drakulix (changes requested):
Good stuff, but duplicating the /api/projects call is not the right way, I think.
This causes a lot of duplicated code, I would like to see the archived status being passed as an query string argument.

Similar to the search function where `limit` and `skip` are optional parameter, I would like to see an `archived` flag added to the existing endpoints, that can either be `true` or `false` and defaults to `false`.

henderkes:
Yeah those should be single methods with ['archived'] as an optional flag like in the manifest schema. This should be easily searchable in ES, too, I guess there should be a way to filter out archived results right in the search call.

backend/api/projects.py
```diff
@@ -166,12 +166,12 @@ def get_project_by_id(project_id):
         res = g.projects.find_one({'_id': project_id})
     else:
         res = None
-    if not res:
+    if res is None:
```
Drakulix: :-1:

Drakulix (dismissed):

backend/api/projects.py
Drakulix: Not is objectively better in most cases.

backend/api/projects.py
```diff
@@ -83,26 +84,36 @@ def add_projects():
...
-        res = g.projects.find({}, limit=limit)
-    elif skip and argc < 2:
-        res = g.projects.find({}, skip=skip)
+    query = {}
+    if archived == 'true':
+        query = {'archived': True}
+    elif archived == 'false':
+        query = {'archived': False}
+    elif archived == 'mixed':
+        query = {}
```
Drakulix: This is actually unnecessary to do, because query is initialized empty anyway.

backend/api/projects.py
```diff
@@ -2,6 +2,7 @@
 Defines API points and starts the application
 """
 
+import sys
```
Drakulix: Could you remove this and the debug prints again?

backend/api/projects.py
```diff
@@ -83,26 +84,36 @@ def add_projects():
...
 
     Returns:
         res: A list of projects
     """
     limit = request.args.get('limit', type=int)
     skip = request.args.get('skip', type=int)
+    archived = request.args.get('archived', type=str, default='false')
+    print(request.query_string, file=sys.stderr)
+
+    print(archived, file=sys.stderr)
```
Drakulix: See above

backend/api/projects.py
```diff
@@ -83,26 +84,36 @@ def add_projects():
...
+    elif archived == 'mixed':
+        query = {}
+    if limit and skip:
+        res = g.projects.find(query, limit=limit, skip=skip)
+    elif limit:
+        res = g.projects.find(query, limit=limit)
+    elif skip:
+        res = g.projects.find(query, skip=skip)
+    elif archived:
+        res = g.projects.find(query)
```
Drakulix: Since archived is set anyway via default, could this not be the final else clause?

backend/api/projects.py
```diff
@@ -83,26 +84,36 @@ def add_projects():
...
     else:
         return make_response('Invalid parameters', 400)
```
Drakulix: See above, can this be reached? If not, it should be removed, as this can be confusing at first.

backend/api/projects.py
```diff
@@ -149,8 +160,16 @@ def get_project_by_id(project_id):
...
-    if not res:
+    archived = request.args.get('archived', type=str, default="false")
+    if archived == 'true':
+        res = g.projects.find_one({'_id': project_id, 'archived': True})
+    elif archived == 'false':
+        res = g.projects.find_one({'_id': project_id, 'archived': False})
+    elif archived == 'mixed':
+        res = g.projects.find_one({'_id': project_id})
+    else:
+        res = None
```
Drakulix: I don't this getting a project directly via Id should require archived to be set, now that we allow everybody to search in all archived projects. Could you revert this part?

backend/api/projects.py
```diff
@@ -218,9 +237,10 @@ def update_project(project_id):
                     '/projects/' + str(manifest['_id']))
                 g.rerun_saved_searches()
                 return make_response("Success")
-            elif not request.on_json_loading_failed():
-                raise ApiException("json could not be parsed",
-                                   400, request.on_json_loading_failed())
+            # remove this so a validation error is thrown
+            # elif not request.on_json_loading_failed:
+            #     raise ApiException("json could not be parsed",
+            #                        400, request.on_json_loading_failed())
```
Drakulix: Just delete this please.

Drakulix (changes requested):
Just rebased (no conflicts) and reviewed this, could you do some small changes quickly @markwitz ? I will then merge this asap.

backend/api/projects.py
markwitz: moved to checking for an allowed status for archived because i's a string to support mixed as an argument

Drakulix (dismissed):

Drakulix (approved):

PR #255, opened 2017-06-29
[BE] permission stuff, added and fixed tests, rewrite upload, refactoring

henderkes:


Drakulix (dismissed):

henderkes:
with @roles_required('admin) -> 404, with @admin_permission.require() -> 500
shouldn't the test return 403 in either case?

henderkes:
Okay this got a little bigger than I thought it would, most issues were so closely related (or the fixes didn't work because other stuff was buggy) that it just resulted in this.

Multiple reviews will be appreciated.

tests/api/test_projects.py
```diff
@@ -348,3 +359,15 @@ def test_invalid_id(self, session, flask_api_url):
...
+        """ Tests for 403 when attempting to update a different users project
+        """
+        data = {"email": "user@knex.com", "password": "user"}
+        session = requests.Session()
+        response = session.post(flask_api_url + '/api/users/login', data=data)
+        assert response.status_code == 200
+        # this should be the id of a project from a different user
+        projectid = str(uuid.uuid4())
+        response = session.delete(flask_api_url + '/api/projects/' + projectid)
+        # assert response.status_code == 403
```
henderkes, line 386: I couldn't figure out how to search for any project where user is not the author quickly (and didn't want to POST one first to use the id) so @JamesComey please edit this later
As of yet a 403 handler for insufficient permission is not implemented.

henderkes:
Note: this is only the start of refactoring. The classes defined there should actually be defined in `backend/api/helpers/usermanagement.py` or something alike

henderkes:
I started refactoring the entire thing but kind of realized that the changes are too severe to be included in this PR and needed more testing. Therefore I reverted the refactoring for the most part and only kept the admin_permission in the new global file.

henderkes:
rebased to current master

Drakulix (approved):

Issue #248, opened 2017-06-29
[BE] test_success_json5_upload


markwitz:
doing nothing on empty array returned

PR #257, opened 2017-06-29
check for json not being empty #248

henderkes:


henderkes:
#248

henderkes:
why does the project upload suddenly fail? I used to verify this by uploading a list of items and got a correct response containing all the ids back then

PR #260, opened 2017-06-30
Provide a decent development environment

Drakulix:
This does a multitude of things:

- It fixes the evalData script to support the new login code
- It adds Docker/compose files to run a development environment that auto-reloads all changes source files upon change.
- It adds a small readme showing how to use it

Everybody may review please.

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -48,4 +48,4 @@ services:
         volumes:
             - ./web:/frontend
         ports:
-            - "3000:3000"
+            - "5000:3000"
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -1,0 +1,51 @@
+version: '3'
+
+services:
+    mongodb:
+        image: mongo
+        command: [ "/usr/bin/mongod", "--replSet", "projectsRepSet" ]
+        ports:
+            - "27017:27017"
+        restart: "on-failure"
+
+    mongodb_replica_setup:
+        build: ./backend/mongodb_replica_setup
+        depends_on:
+            - "mongodb"
+
+    elasticsearch:
+        image: docker.elastic.co/elasticsearch/elasticsearch:5.4.1
+        environment:
+            - "http.host=0.0.0.0"
+            - "transport.host=127.0.0.1"
+            - "xpack.security.enabled=false"
+            - "action.auto_create_index=true"
+        ports:
+            - "9200:9200"
+        restart: "on-failure"
+
+    mongo-connector:
+        build: ./backend/mongo_connector
+        restart: "always"
+        depends_on:
+            - "elasticsearch"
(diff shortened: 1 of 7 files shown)


Drakulix:
Wanted to do some smaller fixes and improvements, but my internet connection is bugging out and I cannot test it, so marking this as WIP. Not yet ready for merge. Still ready for testing and review.

docker-compose-dev.yml
```diff
@@ -0,0 +1,51 @@
...
+        build:
+            context: ./
+            dockerfile: ./Dockerfile-dev-frontend
+        restart: "always"
+        depends_on:
+            - "flask"
+        volumes:
+            - ./web:/frontend
+        ports:
+            - "3000:3000"
```
henderkes: I like this much better than before

evalData/testdata_insertion.py
```diff
@@ -1,23 +1,25 @@
...
 import io
 import os
 import json
+import requests
 
 if __name__ == "__main__":
-    conn = http.client.HTTPConnection("localhost", 5000)
+    session = requests.Session()
+    data = {"email": "admin@knex.com", "password": "admin"}
+    session.post("http://localhost:3000/api/users/login", data=data)
```
henderkes: should these requests really go through the frontend?

gitmirgut:
It works like a charm :tada: (for FE)

evalData/testdata_insertion.py
Drakulix: I don't export the backend port and frontend does only proxy it, so why not.

docker-compose-dev.yml
Drakulix: Better then what? The deployment file is still there (for deployment).

henderkes: Well it's hard to correctly mark the part I meant so it kind of landed in the nowhere. What I meant is also specifying the FE part in the compose yaml and mounting the volumes for instant code changes. We already had that in the BE once and it was very convenient. Also having a shared port to access the through the FE specifically, just better overall.

evalData/testdata_insertion.py
henderkes: Oh, missed that part (BE no longer being exported on port 5000), perfectly makes sense then.

Drakulix:
Ok. I think this is good now.

@DubbleClick @gitmirgut A last test and review before merge?

henderkes:
I think travis doesn't agree

Drakulix:
Okay now hopefully. @DubbleClick

henderkes (dismissed):
looks good now

henderkes:
@gitmirgut  @Rintel or anyone from frontend?

henderkes (dismissed):
taking too long, aproving and merging this now

PR #261, opened 2017-06-30
Profilepage and auth improvements

JamesComey:
Profileview and more auth stuff. Please approve asap, so people can use it. PW change and profile change does not work because of backend issues, from frontend side it should be gud...

### DIFF
--- a/backend/api/users.py
+++ b/backend/api/users.py
@@ -30,7 +30,6 @@ def logout():
 
 
 @users.route('/api/users', methods=['POST'])
-@roles_required('admin')
 def create_user():
     try:
         user = request.get_json()
@@ -57,7 +56,7 @@ def create_user():
 
 
 @users.route('/api/users', methods=['PUT'])
-@roles_required('admin')
+@login_required
 def update_user():
     user = request.get_json()
 
@@ -76,7 +75,7 @@ def update_user():
 
 
 @users.route('/api/users/password', methods=['PUT'])
-@roles_required('admin')
+@login_required
 def update_password():
     user = request.get_json()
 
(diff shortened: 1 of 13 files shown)


backend/api/users.py
```diff
@@ -30,7 +30,6 @@ def logout():
 
 
 @users.route('/api/users', methods=['POST'])
-@roles_required('admin')
```
Drakulix, line 33: Please change this back until we have decided upon this in the large group.

backend/api/users.py
```diff
@@ -57,7 +56,7 @@ def create_user():
 
 
 @users.route('/api/users', methods=['PUT'])
-@roles_required('admin')
+@login_required
```
Drakulix, line 59: I am willing to merge this, but please open an issue then, because you should only be able to update your own profile and that is currently not enforced.

backend/api/users.py
```diff
@@ -76,7 +75,7 @@ def update_user():
 
 
 @users.route('/api/users/password', methods=['PUT'])
-@roles_required('admin')
+@login_required
```
Drakulix, line 78: Same here, see above.

Drakulix (changes requested):
Backend part looks almost good. Anybody may do that small change, but I  am not anymore on my computer today.

henderkes:
Changing these was fairly pointless as it's currently being worked on. The @roles.required('admin') currently does nothing in Blueprints anyways.

backend/api/users.py
henderkes: @roles_required('admin') currently does nothing in blueprints anyways so the change was a bit pointless. it was now changed to @admin_permission.require() in the latest master, which should equally not be removed from master until decided otherwise by everyone together

Drakulix:
Needs to be rebased anyway.

PR #262, opened 2017-06-30
Backend/bugfix/update/iss229

pdler:
added method that checks if a user has admin role. He can change a user (with user role) password without using the old password (maybe the user forgot his password)....


backend/api/helper/userpermission.py
```diff
@@ -0,0 +1,11 @@
+def is_permitted(roles):
+    """Return boolean value if user has admin permission, arg->list with roles
+
+        Returns:
+            res: true if user has admin role
+        """
+    for role in roles:
+        if(role.name == 'admin'):
```
henderkes: we have @roles_required('admin') and @ADMIN_PERMISSION.require() for this

backend/api/helper/userpermission.py
```diff
@@ -0,0 +1,11 @@
+def is_permitted(roles):
```
henderkes: this method also needs the current user and the instruction it should compare the eligiblility to execute on

backend/api/users.py
```diff
@@ -55,26 +58,65 @@ def create_user():
...
-    res.password = encrypt_password(user['password'])
-    res.bio = user['bio']
-    res.roles = []
-    res.roles.append(g.user_datastore.find_or_create_role(user['role']))
-    res.save()
-    res = make_response(dumps(res))
-    res.headers['Content-Type'] = 'application/json'
-
-    return make_response("User with email: " + user['email'] + " updated", 200)
+    is_same_user = editor['email'] == user['email']
```
henderkes: feel like this should be in the is_permitted method

backend/api/users.py
```diff
@@ -1,8 +1,10 @@
 from flask import request, jsonify, make_response, current_app, g, Blueprint
-from flask_security import login_required, roles_required, login_user, logout_user
+from flask_security import login_required, roles_required, login_user,\
+    logout_user, current_user
 from flask_security.utils import verify_password, encrypt_password
-
+from bson.json_util import dumps
```
henderkes: ?

backend/api/users.py
```diff
@@ -1,8 +1,10 @@
 from flask import request, jsonify, make_response, current_app, g, Blueprint
-from flask_security import login_required, roles_required, login_user, logout_user
+from flask_security import login_required, roles_required, login_user,\
+    logout_user, current_user
```
henderkes: ?

backend/api/users.py
```diff
@@ -55,26 +58,65 @@ def create_user():
...
+        res.save()
+        return make_response("Password restored!", 200)
+
+    elif (editor["email"] == user['email']):
+        res = g.user_datastore.get_user(user['email'])
+        if res is None:
+            return make_response("Unknown User with Email-address: " +
+                                 user['email'], 404)
+        old_password = user["old password"]
+        if verify_password(old_password, res.password):
```
henderkes: this check should not be needed if the current user is an admin. he could as well just create a different admin user, log in with that one and then change the password of his old user. doesn't make sense, really.

henderkes (changes requested):

pdler:
I have no time to change this. I am in the frontend team (since wednesday) and I have to learn react for finishing the frontend tasks... So you can do the changes yourself or someone else.

backend/__init__.py
```diff
@@ -6,7 +6,8 @@
 from flask_cors import CORS
 from flask_login import LoginManager
 from flask_mongoengine import MongoEngine
-from flask_security import Security, MongoEngineUserDatastore, UserMixin, RoleMixin
+from flask_security import Security, MongoEngineUserDatastore,\
+    UserMixin, RoleMixin, current_user
```
henderkes: -

backend/api/users.py
```diff
@@ -2,11 +2,22 @@
 from flask_security import login_required, roles_required, login_user,\
     logout_user, current_user
 from flask_security.utils import verify_password, encrypt_password
-
+from bson.json_util import dumps
```
henderkes: -

backend/api/users.py
```diff
@@ -24,29 +35,36 @@ def login():
 
 
 @users.route('/api/users/logout')
+@login_required
```
henderkes: -

backend/api/users.py
```diff
@@ -24,29 +35,36 @@ def login():
...
     return make_response("Logged out", 200)
 
 
 @users.route('/api/users', methods=['POST'])
-@roles_required('admin')
 def create_user():
     try:
         user = request.get_json()
 
+        if user['roles'] == 'admin':
```
henderkes, line 48: not very nice but I guess we won't let the frontend change that

backend/api/users.py
```diff
@@ -24,29 +35,36 @@ def login():
...
+        roles = g.user_datastore.find_or_create_role(user['roles'])
         # if res is not None:
         # return make_response('User already exists',500)
 
         g.user_datastore.create_user(first_name=user['first name'],
                                      last_name=user['last name'],
                                      email=user['email'],
-                                     password=encrypt_password(user['password']),
+                                     password=encrypt_password(
+                                         user['password']),
```
henderkes, line 64: pep8

backend/api/users.py
```diff
@@ -57,41 +75,59 @@ def create_user():
...
+    if(is_permitted(current_user, user)):
+        res = g.user_datastore.get_user(user['email'])
+        if res is None:
+            return make_response("Unknown User with Email-address: " +
+                                 user['email'], 400)
+        res.first_name = user['first name']
+        res.last_name = user['last name']
+        res.bio = user['bio']
+        res.save()
+        res = make_response(dumps(res))
```
henderkes: why do we need dumps here, use jsonify

backend/api/users.py
```diff
@@ -57,41 +75,59 @@ def create_user():
...
+        res.headers['Content-Type'] = 'application/json'
```
henderkes: json5?

backend/api/users.py
```diff
@@ -57,41 +75,59 @@ def create_user():
...
-    res.last_name = user['last name']
-    res.bio = user['bio']
-    res.save()
-    res = make_response(dumps(res))
-    res.headers['Content-Type'] = 'application/json'
+        return make_response("User with email: " +
+                             user['email'] + " updated", 200)
 
-    return make_response("User with email: " + user['email'] + " updated", 200)
+    return make_response("You don't have the permissions " +
```
henderkes: You don't have permission to edit this user.

backend/api/users.py
henderkes: this still holds true, make it ```if current_user.has_role('admin') or verify_password(...)```

backend/api/users.py
```diff
@@ -57,41 +75,59 @@ def create_user():
...
+        if verify_password(old_password, res.password):
+            new_password = user["new password"]
+            if new_password == old_password:
+                return make_response("The old and new passwords" +
+                                     "can not be the same", 200)
+            res.password = encrypt_password(new_password)
+            res.save()
+            return make_response("Password updated!", 200)
+        return make_response("Old password is wrong", 400)
+    if current_user.has_role('admin'):
```
henderkes: I would rather not write the same code twice but just solve it like I suggested above

backend/api/users.py
```diff
@@ -57,41 +75,59 @@ def create_user():
...
+                                 user['email'], 404)
         new_password = user["new password"]
         res.password = encrypt_password(new_password)
         res.save()
-        return make_response("Password updated!", 200)
-    return make_response("Password wrong", 400)
-
+        return make_response("Password restored!", 200)
+    else:
+        return make_response("You don't have the permissions " +
```
henderkes: You don't have permission to edit this user.

henderkes (changes requested):

PR #266, opened 2017-07-01
Feature: Comments

henderkes:
…usermanagement is missing, will be done soon

### DIFF
--- a/backend/api/projects.py
+++ b/backend/api/projects.py
@@ -263,7 +263,7 @@ def add_comment(project_id):
                   or 404 if project is not found
     """
     try:
-        manifest = g.projects.find({'_id': project_id}, projection='comments').sort(
+        manifest = g.projects.find({'_id': project_id}, {'comments': 1, '_id': 0}).sort(
                    key=lambda x: x['datetime'], reverse=True)
         if manifest is None:
             raise ApiException("Project not found", 404)
--- a/backend/api/projects.py
+++ b/backend/api/projects.py
@@ -200,3 +200,190 @@ def update_project(project_id):
                            "the request body does not appear to be utf-8", 400)
     except Exception as err:
         raise ApiException(str(err), 500)
+
+
+@projects.route('/api/projects/<uuid:project_id>/comment', methods=['PUT'])
+@login_required
+def add_comment(project_id):
+    """Adds new comment to project
+
+    Args:
+        project_id, comment in string format
+
+    Returns:
+        response: Success response
+                  or 404 if project is not found
+                  or 400 if validation error occurs
+    """
+    try:
+        manifest = g.projects.find_one({'_id': project_id})
+        if manifest is None:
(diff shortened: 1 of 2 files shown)


backend/api/projects.py
```diff
@@ -187,3 +187,190 @@ def update_project(project_id):
...
+        response: Success response
+                  or 404 if project is not found
+                  or 400 if validation error occurs
+    """
+    try:
+        manifest = g.projects.find_one({'_id': project_id})
+        if manifest is None:
+            raise ApiException("Project not found", 404)
+        if "text/plain" not in request.content_type:
+            raise ApiException("Comment had not format string", 400)
```
Drakulix: Wording seems to be a bit off here.

backend/api/projects.py
```diff
@@ -187,3 +187,190 @@ def update_project(project_id):
...
+        comment = {}
+        comment[author] = None  # How to get this
```
Drakulix: `from flask_security import current_user`

Drakulix (commented):
Already very good.

backend/api/projects.py
henderkes: yes I'm already working on it and tweaking a few things + implementing current user

backend/manifest_schema.json
```diff
@@ -152,6 +152,21 @@
...
+        "type": "array",
+        "items": {
+          "author": {
+            "$ref": "#/project/definitions/user"
+          },
+          "commentId": {
+            "type": "string",
+            "format": "uuid"
+          },
+          "commentMsg": {
```
henderkes: shouldnt this be "comment_id" and "comment_msg" to follow our naming?

Drakulix: You are right, I would even prefer this to be named `id` and `message`.

henderkes:
woops, defined the method before the User class

henderkes:
Do we maybe have to define one field of the user as primary=True?

henderkes:
And apparentpy I forgot pushing my last changes, email should be unique, too.

backend/api/projects.py
```diff
@@ -248,11 +248,11 @@ def add_comment(project_id):
...
     """
     try:
-        manifest = g.projects.find({'_id': project_id}, {'comments': 1, '_id': 0}).sort(
+        comments = g.projects.find({'_id': project_id}, {'comments': 1, '_id': 0}).sort(
                    key=lambda x: x['datetime'], reverse=True)
-        if manifest is None:
+        if comments is None:
             raise ApiException("Project not found", 404)
-        return jsonify(manifest['comments'])
+        return jsonify(comments)
```
henderkes: I dont think this is the intended return. The (list of) comments should be returned, not the entire json I think?

Drakulix:
@DubbleClick Is this now ready for merge except for this one nitpick?

PR #272, opened 2017-07-03
[FE] Add material icons support

Rintel:
# What? 
This adds support for googles material icons. 
You can find the reference [here](https://material.io/icons/).


mharasic (approved):
New Icons!

PR #276, opened 2017-07-05
Tests for Search, and for consistent delete

Leats:
~~@app.route('/api/projects/search', methods=['POST'])~~
~~Doesn't seem to work for me, unless I made a mistake with the queries?~~
EDIT: Since that isn't used at all I threw those tests out again.

GET doesn't make any problems, though, afaik.

The delete is probably a bit not at all elegant, if anyone has a better way to find the container, I'd be grateful. It does seem to work, though.

Sorry about the many commits. It's not working on my computer and I have kinda given up and instead work now by letting travis test it.

### DIFF
--- a/tests/api/test_database_connection.py
+++ b/tests/api/test_database_connection.py
@@ -55,7 +55,7 @@ def test_consistent_delete_no_elastic(session, flask_api_url, pytestconfig, elas
 
     container.start()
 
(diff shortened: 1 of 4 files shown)


henderkes:
I feel like this isn't worth rewriting, so #closed

PR #281, opened 2017-07-05
saved searches & notification fixes

Drakulix:
## What

- This does add saved searches to the user object
- Changes the search endpoints to allow saving
- Creates endpoint to list searches, delete search and query a search again
- Reruns all searches when projects do change
- Create notification
- Fixup notifications
- Add notification limit & deduplication by link
- Setup routing for the results of the saved search

## Missing

- Adding functionality to the frontend for actually saving
- Adding functionality to the frontend for listing searches in the profile


mharasic (approved):
looks good

Issue #250, opened 2017-06-29
[BE] Add endpoint to get authors


Rintel:

Drakulix:
This is needed for auto-completion and should contain all authors known from the projects already deduplicated.

Should be easy to do with [`distinct`](https://docs.mongodb.com/manual/reference/method/db.collection.distinct/)

henderkes:
done in #286

Issue #283, opened 2017-07-06
[BE] update_project function does not work


pdler:
- current_user is not defined
   - please use current_user from flask_security

henderkes:
see #286

Issue #279, opened 2017-07-05
[BE] Filter get user call


Drakulix:
Replaces roles with names, remove password and others.

Issue #270, opened 2017-07-02
[BE] User API getUser


pdler:
The json file only gives the roles as id value. But we need the roles as String

henderkes:
Can you elaborate a bit? The /api/users GET call?

pdler:
Yes, we only have a call which gives us a user with given email address. A call that gives us all user does still not exist or? The /api/users GET call gives a json which contains the roles as id. The frontend does not knows which id means an admin or user? So it will be better if the roles field contains the roles as string instead of an id.

Btw: the json contains the hashed password. it will be better if you remove this field in the /api/users GET call

henderkes:
Alright, I have no idea how to get the corresponding strings attached to that role (not sure if they even exist by default) but I will update once this is fixed

Drakulix:
@DubbleClick I think this is only of interested to know, if the user is an admin or not ( right? @pdler ).
So if we can match that and just add a field called "is_admin" or something like that, that should be enough.

henderkes:
I thought about it a bit and strongly consider implementing our own permission checks and just completely forgetting about Flasks stuff except for the login. That way we can define id and roles to our liking and aren't bound to ReferenceFields with no documentation. I might try that later and if it gets too complicated I'll just define a secondary ListField role_names.

PR #286, opened 2017-07-06
[WIP] Fixed comments, user GET, refactoring, pylint

Krisselchen:
fixed message field in manifest.
Furthermore fixed some creepy bugs in comments.
I am not happy with line 268 to 270: 
The only solution I had, would be, to initialize comments in projects POST for every manifest with an empty array in comments field to distinguish between no projects found an no comments found. Of course comments have to be required in scheme in this case. Any different ideas?

### DIFF
--- a/backend/api/projects.py
+++ b/backend/api/projects.py
@@ -166,6 +166,8 @@ def update_project(project_id):
         elif request.is_json or "application/json5" in request.content_type:
             manifest = request.get_json() if request.is_json\
                 else json5.loads(request.data.decode("utf-8"))
+            if manifest['id'] != str(project_id):
+                return make_response("project_id and json['id'] do not match.", 409)
             is_valid = g.validator.is_valid(manifest)
             if is_valid and is_permitted(current_user, manifest):
                 manifest['_id'] = project_id
--- a/backend/api/projects.py
+++ b/backend/api/projects.py
@@ -255,7 +255,8 @@ def get_comment(project_id):
         project = g.projects.find_one({'_id': project_id})
         if not project:
             raise ApiException("Project not found", 404)
-        res = project['comments'] if project['comments'] else []
+        res = sorted(project['comments'], key=lambda k: k['datetime'],
+                     reverse=True) if project['comments'] else []
         return jsonify(res)
     except ApiException as error:
         raise error
--- a/backend/api/projects.py
+++ b/backend/api/projects.py
@@ -297,8 +297,10 @@ def delete_comments(project_id):
             raise ApiException("Project not found", 404)
         if 'comments' in manifest:
             del manifest['comments']
+            manifest['_id'] = str(project_id)
             is_valid = g.validator.is_valid(manifest)
             if is_valid:
+                manifest['_id'] = project_id
                 g.projects.find_one_and_replace({'_id': project_id}, manifest,
                                                 return_document=ReturnDocument.AFTER)
                 return make_response("Success", 200)
@@ -344,8 +346,10 @@ def update_comment(project_id, comment_id):
                 if not is_permitted(current_user, comment):
                     raise ApiException("Permission denied", 403)
                 comment['message'] = request.data.decode('utf-8')
+                manifest['_id'] = str(project_id)
                 is_valid = g.validator.is_valid(manifest)
                 if is_valid:
+                    manifest['_id'] = project_id
                     g.projects.find_one_and_replace({'_id': project_id}, manifest,
                                                     return_document=ReturnDocument.AFTER)
                     return make_response("Success", 200)
@@ -396,8 +400,10 @@ def delete_comment(project_id, comment_id):
                 manifest['comments'].remove(comment)
                 if not manifest['comments']:
                     del manifest['comments']
+                manifest['_id'] = str(project_id)
                 is_valid = g.validator.is_valid(manifest)
                 if is_valid:
+                    manifest['_id'] = project_id
                     g.projects.find_one_and_replace({'_id': project_id}, manifest,
                                                     return_document=ReturnDocument.AFTER)
                     return make_response("Success", 200)
--- a/backend/api/projects.py
+++ b/backend/api/projects.py
@@ -257,9 +257,9 @@ def get_comment(project_id):
         project = g.projects.find_one({'_id': project_id})
         if not project:
             raise ApiException("Project not found", 404)
-        res = sorted(project['comments'], key=lambda k: k['datetime'],
-                     reverse=True) if project['comments'] else []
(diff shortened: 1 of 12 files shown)


backend/api/projects.py
```diff
@@ -327,18 +310,17 @@ def update_comment(project_id, comment_id):
...
 
         if not manifest['comments']:
             raise ApiException("Project has no comments", 404)
-        # TODO Usermanagement
-        comment_msg = request.data.decode('utf-8')
+
         for comment in manifest['comments']:
-            if str(comment_id) == str(comment['id']):
-                comment['message'] = comment_msg
+            if str(comment_id) == str(comment['id']) and is_permitted(current_user, comment):
```
henderkes: shouldn't a user be able to delete or edit any comments under his projects, even if they aren't his own?

henderkes: @Krisselchen answered this, it's working as intended here

henderkes (dismissed):
approved from me, please wait for a second approval because many changes were from me too

henderkes:
Mobile github site is the best...

henderkes:
this fixes #250, #251, #265, #270, #279, #283, #290

yeah.. a bit much for one pr, I'm sorry

gitmirgut:
`POST` '/api/projects/<uuid:project_id>/comment' not working.
message:
```
{
    "message": "name 'user' is not defined",
    "validation_error": null
}
```

henderkes:
hotfixed it, find&replace for user->current_user didn't work because there was no space before it after a + operator

gitmirgut:
POST '/api/projects/uuid:project_id/comment' not still working.
message:
```
{
    "message": "maximum recursion depth exceeded in comparison",
    "validation_error": null
}
```

backend/api/projects.py
```diff
@@ -297,20 +293,22 @@ def delete_comments(project_id):
...
-            make_response("Success")
-        else:
-            validation_errs = [error for error in
-                               sorted(g.validator.iter_errors(manifest))]
-            if validation_errs is not None:
-                raise ApiException("Validation Error: \n" +
-                                   str(is_valid), 400, validation_errs)
+        if 'comments' in manifest:
+            del manifest['comments']
+            manifest['_id'] = str(project_id)
```
henderkes, line 321: we should adjust the validator long term, this is a poor solution but works for now

backend/api/projects.py
```diff
@@ -368,48 +372,46 @@ def update_comment(project_id, comment_id):
...
+    """Deletes comment by comment_id
 
     Args:
         project_id, comment_id
 
     Returns:
         response: Success response
                   or 404 if project is not found
-                  or 400 if validation error occurs
+                  or 500 if validation error occurs
```
henderkes, line 404: I think a validation error shouldn't return 500

backend/api/search.py
```diff
@@ -1,10 +1,25 @@
...
+def get_all_authors():
+    authors = g.projects.distinct('authors')
+    return jsonify(authors)
+
+
+@search.route('/api/search/tags', methods=['GET'])
+@login_required
+def get_all_tags():
+    tags = g.projects.disctinct('tags')
+    return jsonify(tags)
```
henderkes: should probably add exception handling

backend/api/users.py
```diff
@@ -40,6 +40,13 @@ def logout():
     return make_response("Logged out", 200)
 
 
+@users.route('/api/users', methods=['GET'])
+@login_required
+def get_all_users():
+    users = g.user_datastore.user_model.objects
+    return jsonify([user.to_dict() for user in users])
```
henderkes, line 47: exception handling too maybe

henderkes (approved):
A couple TODOs, approved otherwise

backend/api/projects.py
Krisselchen: But the validation Error occurs by the backend. The projectmanifest is only used internally here. I am not really sure about it.

backend/api/search.py
Krisselchen: Shouldn't it be in projects.py?

henderkes: I'm not sure, felt like it's more of a search thing but we can move it to projects

backend/api/users.py
```diff
@@ -147,29 +156,18 @@ def insert_bookmarks(id):
...
-        return make_response("Unknown User with Email-address: " +
-                             user['email'], 400)
-
+    res = g.user_datastore.get_user(current_user['email'])
     if id in res.bookmarks:
         res.bookmarks.remove(id)
         res.save()
         return jsonify(res['bookmarks'])
-    return make_response("Project is not bookmarked: ", 400)
+    return make_response("Project is not bookmarked: " + id, 400)
```
henderkes: needs an explicit str(id) cast

henderkes (changes requested):

backend/api/projects.py
```diff
@@ -108,6 +108,30 @@ def get_projects():
...
 
+@projects.route('/api/projects/authors', methods=['GET'])
+@login_required
+def get_all_authors():
+    try:
+        authors = g.projects.distinct('authors')
+        if not authors:
+            return jsonify([])
+        result = sorted([author['name'] for author in authors] +
+                        [author['email'] for author in authors], key=str.lower)
```
henderkes: shouldn't we return a list of author dicts instead? I mean we can sort them this way, but returning names first and emails after with no real distinction doesn't seem a good idea

henderkes (changes requested):

henderkes (dismissed):

backend/api/users.py
mharasic: with cast its not working unfortunately

henderkes: are you sure? it's not working without a cast because it can't implicitly convert uuid to str, but explicit conversion from uuid to str works just fine

Drakulix: What means not working?

backend/api/projects.py
henderkes: this is the DELETE call so I was mistaken in the first place, 500 is correct and PUT and POST were already returning correct codes

backend/api/users.py
henderkes: it raised a 500 error because flask didn't handle the exception (because it can't implitly cast). 

changed it to str(id) now and I don't see a reason why it wouldn't work now, it's only the response text anyways.

henderkes (dismissed):
More reviews please.

Drakulix (approved):

henderkes (approved):

Leats (approved):

PR #289, opened 2017-07-08
merge this quick please, two lines changed

henderkes:


henderkes:
braces were missing and there is no variable "last", which should instead of the end delimiter

Krisselchen (approved):

PR #301, opened 2017-07-10
Add update and comment notifications

Drakulix:
Somebody an idea how I can get pep8 happy here? autopep8 is not able to fix it.


henderkes:
Well four lines are too long, so either line break them after braces or manually escape the line with a backslash

henderkes (approved):

PR #305, opened 2017-07-12
[FE] Code Formating

Rintel:


Rintel:
Does not make sense anymore to refactor as too much stuff has changed in between. Closed.

PR #312, opened 2017-07-13
Backend/security test

aTepliak:
this is   updated  security test (login etc), update of  https://github.com/Drakulix/knex/tree/backend/test/security/iss165


tests/api/test_security.py
```diff
@@ -0,0 +1,80 @@
+import requests
+import unittest
+import os
+
+
+class Testsecurity(object):
+    def setup(self):
```
markwitz: setup does nothing?

tests/api/test_security.py
```diff
@@ -0,0 +1,80 @@
...
+        # app['TESTING'] = True
+        # self.app = app.test_client()
+        pass
+
+    def test_main_page(self, flask_api_url):
+        response = requests.get(flask_api_url + '/')
+        assert response.status_code == 200
+
+    def test_login_fake_user(self, flask_api_url):
+        response = requests.post(flask_api_url + '/api/users/login',
```
markwitz: these tests have to make sure that the user used is really in the database so maybe create and delete them before and after the test. But this might be implemented when we implement the the config file

tests/api/test_security.py
```diff
@@ -0,0 +1,80 @@
+import requests
+import unittest
```
markwitz: do we need unittests?

markwitz (dismissed):
overall good work

tests/api/test_security.py
aTepliak: These are default users 
They're are created in init 
I can create a new user,  and delete it afterwards. Is mongoengine a database?

tests/api/test_security.py
aTepliak: corrected

tests/api/test_security.py
aTepliak: deleted

tests/api/test_security.py
aTepliak: or alternatively i can check if user is in mongoengine or not, but the initial idea was just to  test wrong  email and password to see if login works properly

henderkes: uh, you could just try deleting the user right before the call, that way it's guaranteed to not be in the database. however, since the tests are mainly for ravis I don't really see why we would need to, it's a clean build with only the default users (user&admin@knex) anyway

tests/api/test_security.py
```diff
@@ -0,0 +1,75 @@
...
+        )
+        with open(test_manifest, 'r', encoding='utf-8') as tf:
+            data = str(tf.read().replace('\n', ''))
+        response = requests.post(flask_api_url + "/api/projects",
+                                 data=data.encode('utf-8'),
+                                 headers={'Content-Type': 'application/json5'})
+
+        assert response.status_code == 200
+
+    def test_access_login_requiered_not_logged(self, pytestconfig, flask_api_url):
```
henderkes: required*

tests/api/test_security.py
```diff
@@ -0,0 +1,75 @@
...
+                                 headers={'Content-Type': 'application/json5'})
+
+        assert response.status_code == 200  # or 500?
+
+    def test_update_user(self, flask_api_url):
+        response = requests.get(flask_api_url + '/api/users/',
+                                data=dict(email='user@knex.com'))
+        assert response.status_code == 404  # 200?
+
+    def test_update_user_not_exists(self, flask_api_url):
```
henderkes: nonexistent instead of not exists please

tests/api/test_security.py
```diff
@@ -0,0 +1,75 @@
...
+        assert response.status_code == 403
+
+    def test_logout(self, flask_api_url):
+        response = requests.get(flask_api_url + '/api/users/login',
+                                data=dict(email='user@knex.com', password="user"))
+        response = requests.get(flask_api_url + '/api/users/logout')
+        assert response.reason == 'OK'
+        assert response.status_code == 200
+
+    def test_access_login_required_logged(self, pytestconfig, flask_api_url):
```
henderkes: this test shouldn't even work, you don't login anywhere. why does this run with assert response.status_code == 200 when it should return 403?

henderkes (changes requested):

henderkes:
@aTepliak ?

tests/api/test_security.py
aTepliak: well it's seem to be a bug; I assume @login_ required doesn't work correctly

tests/api/test_security.py
```diff
@@ -62,14 +68,14 @@ def test_access_login_requiered_not_logged(self, pytestconfig, flask_api_url):
...
         response = requests.get(flask_api_url + '/api/users/',
                                 data=dict(email='user@knex.com'))
         assert response.status_code == 404  # 200?
 
-    def test_update_user_not_exists(self, flask_api_url):
+    def test_update_user_nonexistent(self, flask_api_url):
         response = requests.get(flask_api_url + '/api/users/',
                                 data=dict(email='unknownuser@knex.com'))
-        assert response.status_code == 404
+        assert response.status_code == 404
```
henderkes: needs a blank line at the end of the file, pep8 will complain otherwise

tests/api/test_security.py
aTepliak: corrected

tests/api/test_security.py
aTepliak: corrected

tests/api/test_security.py
aTepliak: corrected

tests/api/test_security.py
```diff
@@ -0,0 +1,81 @@
+import requests
+import unittest
```
henderkes: This import is actually useless as t's never used, unless I am missing something?

tests/api/test_security.py
```diff
@@ -0,0 +1,81 @@
...
+import os
+
+
+class Testsecurity(object):
```
henderkes: This class should be called TestPOST or alike, depending on which methods are tested.

tests/api/test_security.py
```diff
@@ -0,0 +1,81 @@
...
+        assert response.status_code == 200
+
+    def test_login_real_user_wrong_psswd(self, flask_api_url):
+        response = requests.post(flask_api_url + '/api/users/login',
+                                 data=dict(email='admin', password='a'))
+        assert response.reason == 'FORBIDDEN'
+        assert response.status_code == 403
+
+    def test_logout(self, flask_api_url):
+        response = requests.get(flask_api_url + '/api/users/login',
```
henderkes: the login is a POST call, you also need to do:

```
session = requests.Session()
session.post(login)
session.get(logout)
```

tests/api/test_security.py
```diff
@@ -0,0 +1,81 @@
...
+        )
+        with open(test_manifest, 'r', encoding='utf-8') as tf:
+            data = str(tf.read().replace('\n', ''))
+        response = requests.post(flask_api_url + "/api/projects",
+                                 data=data.encode('utf-8'),
+                                 headers={'Content-Type': 'application/json5'})
+
+        assert response.status_code == 403  # or 500?
+
+    def test_update_user(self, flask_api_url):
```
henderkes: this doesn't test updating a user but getting one

henderkes (changes requested):

tests/api/test_security.py
```diff
@@ -1,45 +1,44 @@
...
+        response = session.post(flask_api_url + '/api/users/login',
                                  data=dict(email='user1', password='password'))
 
         assert response.reason == 'FORBIDDEN'
         assert response.status_code == 403
 
-    def test_login_real_user(self, flask_api_url):
-        response = requests.post(flask_api_url + '/api/users/login',
+    def test_login_real_user(self,session, flask_api_url):
+        response = session.post(flask_api_url + '/api/users/login',
```
henderkes: session already logs you in in the fixture. Please actually do session = requests.Session() in the method if you want to test login/out

henderkes (changes requested):

tests/api/test_security.py
```diff
@@ -13,6 +13,7 @@ def test_main_page(self, session, flask_api_url):
         assert response.status_code == 200
 
     def test_login_fake_user(self,session, flask_api_url):
+        session = requests.Session()
```
henderkes: it's not only this method and you shouldn't get the conftest.py fixture as method argument then

henderkes (changes requested):

tests/api/test_security.py
aTepliak: I added it to every method and added login after, / one time tried to reinitialize session variable, but it doesn't work anyway

henderkes: Yes, but you shouldn't use the conftest fixture "session" as a method argument. You create your own session, there is no need to get that of the fixture.

tests/api/test_security.py
henderkes: this is still the case. please do not get the session fixture if the purpose of the method is testing the login. Instead create your own session as above. **Without** the fixture.

tests/api/test_security.py
```diff
@@ -65,18 +65,18 @@ def test_access_login_required_not_logged(self, session, pytestconfig, flask_api
...
         assert response.status_code == 404  # 200?
 
-    def test_update_user_nonexistent(self,session, flask_api_url):
+    def test_update_user_nonexistent(self, session, flask_api_url):
         response = session.get(flask_api_url + '/api/users/',
-                                data=dict(email='unknownuser@knex.com'))
+                               data=dict(email='unknownuser@knex.com'))
         assert response.status_code == 404
-
+        session = requests.Session()
```
henderkes: ??

henderkes (changes requested):

tests/api/test_security.py
```diff
@@ -0,0 +1,85 @@
+import requests
+import os
+
+
+class TestPOST(object):
+    def setup(self):
+        # app['TESTING'] = True
+        # self.app = app.test_client()
+        pass
```
henderkes: can't we just delete this if it isn't used?

tests/api/test_security.py
```diff
@@ -0,0 +1,85 @@
...
+    def test_main_page(self, flask_api_url):
+        response = requests.get(flask_api_url + '/')
+        assert response.status_code == 200
+
+    def test_login_fake_user(self, flask_api_url):
+        session = requests.Session()
+        response = session.post(flask_api_url + '/api/users/login',
+                                data=dict(email='user1', password='password'))
+
+        assert response.reason == 'FORBIDDEN'
```
henderkes: as mentioned above, checking response reasons is pretty pointless if you already check status codes.

tests/api/test_security.py
```diff
@@ -0,0 +1,85 @@
...
+        assert response.status_code == 403
+
+    def test_login_real_user(self, flask_api_url):
+        session = requests.Session()
+        response = session.post(flask_api_url + '/api/users/login',
+                                data=dict(email='admin@knex.com', password="admin"))
+        assert response.reason == 'OK'
```
henderkes: see above

tests/api/test_security.py
```diff
@@ -0,0 +1,85 @@
...
+        assert response.status_code == 200
+
+    def test_login_real_user_wrong_psswd(self, flask_api_url):
+        session = requests.Session()
+        response = session.post(flask_api_url + '/api/users/login',
+                                data=dict(email='admin', password='a'))
+        assert response.reason == 'FORBIDDEN'
```
henderkes: see above

tests/api/test_security.py
```diff
@@ -0,0 +1,85 @@
...
+        assert response.status_code == 403
+
+    def test_login_real_user(self, flask_api_url):
+        session = requests.Session()
+        response = session.post(flask_api_url + '/api/users/login',
+                                data=dict(email='admin@knex.com', password="admin"))
+        assert response.reason == 'OK'
+        assert response.status_code == 200
+
+    def test_login_real_user_wrong_psswd(self, flask_api_url):
```
henderkes: can we rename this to test_login_wrong_password?

tests/api/test_security.py
```diff
@@ -0,0 +1,85 @@
...
+                                data=dict(email='admin', password='a'))
+        assert response.reason == 'FORBIDDEN'
+        assert response.status_code == 403
+
+    def test_logout(self, flask_api_url, session):
+        session = requests.Session()
+        response = session.get(flask_api_url + '/api/users/login',
+                               data=dict(email='user@knex.com', password="user"))
+        response = session.get(flask_api_url + '/api/users/logout')
+        assert response.reason == 'OK'
```
henderkes: see above

tests/api/test_security.py
```diff
@@ -0,0 +1,85 @@
...
+        session = requests.Session()
+        response = session.post(flask_api_url + '/api/users/login',
+                                data=dict(email='admin', password='a'))
+        assert response.reason == 'FORBIDDEN'
+        assert response.status_code == 403
+
+    def test_logout(self, flask_api_url, session):
+        session = requests.Session()
+        response = session.get(flask_api_url + '/api/users/login',
+                               data=dict(email='user@knex.com', password="user"))
```
henderkes: I don't think a login GET call works

tests/api/test_security.py
```diff
@@ -0,0 +1,85 @@
...
+                                 headers={'Content-Type': 'application/json5'})
+
+        assert response.status_code == 403  # or 500?
+
+    def test_get_user(self, session, flask_api_url):
+        response = session.get(flask_api_url + '/api/users/',
+                               data=dict(email='user@knex.com'))
+        assert response.status_code == 404  # 200?
+
+    def test_update_user_nonexistent(self, flask_api_url):
```
henderkes: this is still not an update user call but a GET user call. it shouldn't be called update as such

tests/api/test_security.py
```diff
@@ -0,0 +1,85 @@
...
+            'testmanifests',
+            'validexample0.json5'
+        )
+        with open(test_manifest, 'r', encoding='utf-8') as tf:
+            data = str(tf.read().replace('\n', ''))
+        response = requests.post(flask_api_url + "/api/projects",
+                                 data=data.encode('utf-8'),
+                                 headers={'Content-Type': 'application/json5'})
+
+        assert response.status_code == 403  # or 500?
```
henderkes: 403 for forbidden is correct. 500 is an internal server error, why would that be desired outcome

henderkes (changes requested):

tests/api/test_security.py
```diff
@@ -0,0 +1,73 @@
...
+        response = requests.post(flask_api_url + "/api/projects",
+                                 data=data.encode('utf-8'),
+                                 headers={'Content-Type': 'application/json5'})
+
+        assert response.status_code == 403
+
+    def test_get_user(self, session, flask_api_url):
+        response = session.get(flask_api_url + '/api/users/',
+                               data=dict(email='user@knex.com'))
+        assert response.status_code == 404
```
henderkes: shouldn't this give a 200? user@knex.com should be in the database

henderkes (changes requested):

tests/api/test_security.py
Drakulix: Yes. @markwitz added a fixture to support this user again, so it is easily fixable.

aTepliak: @markwitz  what the fixture is it?

aTepliak: @markwitz  which fixture is it?

henderkes: look in conftest.py

henderkes (approved):

PR #313, opened 2017-07-13
[WIP] Backend/feature/avatar

henderkes:


henderkes:
this forces us to install an imaging library, I just chose pillow as default

henderkes:
I'm not sure why it doesn't find the file, maybe because __path__ is set to that of __init__,py? I'm very unfamiliar with pythons relative path system

henderkes:
@mharasic by default a user gets the default image, put calls update (replace) the image and DELETE calls don't actually delete the image, but reassign the default

the avatar is not sent when using /api/users GET, only with /api/users/<email>/avatar
To upload an image do the same as in project upload by file, but with "image" instead of "files[]".
Content-Type header must be set to "image/png" or "image/jpeg", etc.

henderkes:
I'm not sure why or with what this is failing anymore

Drakulix (approved):

PR #316, opened 2017-07-14
[BE] saved search metadata

Drakulix:
- Removes unused search apis
- Makes the saved search delete call more consistent with the other delete calls
- Replaces the title field with a metadata field, that can be freely set by the frontend, but is required to be valid json. It's a little dirty to save it internal as a string, but that works around having to specify the concrete and possibly quite complex or even dynamic type of the json.


backend/api/search.py
```diff
@@ -44,7 +44,7 @@ def search_simple():
     count = request.args.get('count', type=int)
     if count is None:
         count = 10
-    save = request.args.get('save', type=str)
+    save = json.loads(request.args.get('save', type=str))
```
henderkes: wouldn't json.loads(None) or json.loads("") throw an exception? should handle the case that 'save' is not submitted

backend/api/search.py
```diff
@@ -271,5 +176,5 @@ def delete_saved_search(id):
         if search.saved_search_id == ObjectId(id):
             user.saved_searches.remove(search)
             user.save()
-            return jsonify([search.to_dict() for search in user.saved_searches])
+            return make_response("Success", 200)
```
henderkes, line 185: Why do we only return a success response for the GET now? This makes no sense.

Edit: Nevermind, this is the delete method, but a GET call?

henderkes (changes requested):

backend/api/search.py
Drakulix: This is clearly a delete call, nothing wrong about it.

backend/api/search.py
Drakulix: you are right. fixed in the latest commit

backend/api/search.py
```diff
@@ -77,7 +77,7 @@ def search_simple():
         raise ApiException(str(e), 400)
 
     if save:
-        return g.save_search(current_user, save, request_json, len(projects))
+        return g.save_search(current_user, json.loads(save), request_json, len(projects))
```
henderkes: this can still throw an exception and lead to a 500 when the string isn't valid json

backend/api/search.py
```diff
@@ -127,106 +127,11 @@ def search_avanced():
...
```

henderkes (changes requested):

henderkes (dismissed):

henderkes (approved):

PR #330, opened 2017-07-17
store uuid in notification link instead of str, fix bookmarks and not…

henderkes:
…ifications on project deletion


mharasic:
:+1:

mharasic (approved):
Sounds good

PR #349, opened 2017-07-19
Backend/config file

### LINKED ISSUE #282: [BE] config file
Before this project is done we need a config file to be able to set the following settings:

- [x] Set mongodb url (& mongodb authentication if possible in the timeframe)
- [x] Set elasticsearch url (& mongodb authentication if possible in the timeframe)
- [x] Set listening interface and port (0.0.0.0:5000 for example)
- [x] Set the 'SECRET_KEY' and 'SECURITY_PASSWORD_SALT' variables
- [x] Set a must-have admin account that is created if not existing (mail & password) from the config file

All settings should default to values, if they are not set or if no config file exists, that match our current docker-setup values. This way our containers continue to run even if no config file is found.

The config file should be in any human-readable format like json,yaml,cfg,etc.

The location to read from should be the current folder of the python folder (so `backend` in our case), but an example config file should be placed into the root of the repository so our container does not load.

markwitz:
resolve #282 

done so far


backend/__init__.py
```diff
@@ -163,9 +196,9 @@ def to_dict(self):
 
 
 class EmailConverter(BaseConverter):
-    regex = r"([a-z0-9!#$%&'*+\/=?^_`{|}~-]+(?:\." +\
-            r"[a-z0-9!#$%&'*+\/=?^_`"r"{|}~-]+)" +\
-            r"*(@|\sat\s)(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?" +\
+    regex = r"([a-z0-9!#$%&'*+\/=?^_`{|}~-]+(?:\." + \
+            r"[a-z0-9!#$%&'*+\/=?^_`"r"{|}~-]+)" + \
+            r"*(@|\sat\s)(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?" + \
```
henderkes: please remove the whitespaces

backend/__init__.py
```diff
@@ -27,18 +28,48 @@
...
+        if "flask" in config:
+            default_config["flask"].update(config["flask"])
+        if "mongo_db" in config:
+            default_config["mongo_db"].update(config["mongo_db"])
+        if "elasticsearch" in config:
+            default_config["elasticsearch"].update(config["elasticsearch"])
+        if "administration_user" in config:
+            default_config["administration_user"].update(config["administration_user"])
+
+config = default_config
```
henderkes, line 69: We should put the default_config in a seperate file as well in my opinion.

backend/__init__.py
```diff
@@ -243,10 +276,11 @@ def on_project_deletion():
...
                               '/project/' not in str(x.link) or g.projects.find_one(
-                               {'_id': uuid.UUID(
-                                   str(x.link)[str(x.link).index('/project/') + len('/project/'):]
-                                   )
-                                })]
+                                  {'_id': uuid.UUID(
+                                      str(x.link)[str(x.link).index('/project/') +
+                                                  len('/project/'):]
+                                  )
+                                  })]
```
henderkes: Feels harder to read than before

tests/api/test_projects.py
```diff
@@ -276,12 +276,13 @@ def test_inconsistent_delete(self, session, flask_api_url, pytestconfig):
...
         assert delete_response.status_code == 200
 
-    def test_unauthorized_delete(self, flask_api_url):
+    def test_unauthorized_delete(self, flask_api_url, enter_default_user_users):
         """ Tests for 404 when attempting to delete a project when
             logged in as a user.
             Not found because the user should not know about the
             /api/delete endpoint.
         """
+        response = enter_default_user_users
```
henderkes, line 285: why? this is not used anywhere

tests/api/test_projects.py
```diff
@@ -505,9 +506,10 @@ def test_invalid_id(self, session, flask_api_url):
     def test_success(self, session, flask_api_url):
         assert True
 
-    def test_unauthorized_update(self, flask_api_url):
+    def test_unauthorized_update(self, flask_api_url, enter_default_user_users):
         """ Tests for 403 when attempting to update a different users project
         """
+        response = enter_default_user_users
```
henderkes, line 513: again, why?

backend/__init__.py
Drakulix: not necessary

backend/__init__.py
```diff
@@ -27,18 +28,48 @@
...
+default_config["mongo_db"]["secret_key"] = "super-secret"
+default_config["mongo_db"]["hostname"] = "mongodb"
+default_config["mongo_db"]["port"] = 27017
+default_config["mongo_db"]["security_password_hash"] = 'pbkdf2_sha512'
+default_config["mongo_db"]["security_password_salt"] = 'THISISMYOWNSALT'
+default_config["elasticsearch"]["hostname"] = "elasticsearch"
+default_config["elasticsearch"]["port"] = 9200
+default_config["administration_user"]["username"] = "admin"
+default_config["administration_user"]["password"] = "admin"
+default_config["administration_user"]["email"] = "admin@knex.com"
```
Drakulix: You could write that a single literal, that would look much nicer.

backend/__init__.py
Drakulix: not necessary in my opinion.

backend/__init__.py
Drakulix: Is that a pep8 thing? Seems your editor @markwitz does that all the time?

backend/__init__.py
Drakulix: Not too important, but @DubbleClick is right.

backend/__init__.py
```diff
@@ -259,18 +293,14 @@ def project_deleted_func():
...
     user_role = USER_DATASTORE.find_or_create_role('user')
     admin_role = USER_DATASTORE.find_or_create_role('admin')
-    userpw = hash_password("user")
-    adminpw = hash_password("admin")
-    try:
-        USER_DATASTORE.create_user(
-            email="user@knex.com", password=userpw, roles=[user_role])
-    except NotUniqueError:
-        pass
+    # userpw = hash_password("user")
```
Drakulix: Please remove, if that is useless now.

tests/api/test_projects.py
Drakulix: The test requires the user to exist, this just runs the fixture.

tests/api/test_projects.py
Drakulix: Same. this is good enough for a test.

Drakulix (changes requested):
Good stuff, minor nitpicks.

tests/api/test_projects.py
markwitz: we could check for the response to be correct here but i just assume it is working correctly

backend/__init__.py
markwitz: i'm at a loss here how do i do that?

Drakulix: Looks like json, but is valid python syntax:
```python
default_config = {
    ...,
    'administration_user': {
        'username': 'admin',
        'password': 'admin',
        'email': 'admin@knex.com',
    }
}
```

henderkes (dismissed):
looks good to me now. formatting in on_project_deletion could still be reverted, but it doesn't matter as long as pep8 is happy

Drakulix (dismissed):

henderkes (approved):

PR #351, opened 2017-07-19
[FE] New Backend/Authentication (detect logout and already logged in)

Drakulix:
## What?

- Rebuild `Backend` as class to include current profile & login state
- Rebuild that state via Cookie and Request upon page load (reloading detects if still logged in or not)
- Handle logout by redirecting to SignIn page
- Model all Backend requests as functions of the Backend class
- Consistent returns as a Promise with value depending on the type of the Backend call (get, put, post, delete)
- Rewrite all components to use the new function, except DataTable, which still uses a manual get for the fetchURL property. (Should probably be replaced by a fetchFunction)

## Please test!


backend/__init__.py
```diff
@@ -158,11 +160,14 @@ class EmailConverter(BaseConverter):
...
 SECURITY = Security(app, USER_DATASTORE)
 
+
 @SECURITY.login_manager.unauthorized_handler
 def handle_unauthorized_access():
     return make_response("Forbidden", 403)
 
 # internal function to append notifications to the given userlist
+
+
```
henderkes: why? one new line before the comment and none after, please

henderkes (dismissed):

backend/__init__.py
Drakulix: grr, that was autopep8, gimme a sec to fix this

mharasic:
comment section is broken and not working
redirection to sign in is not working if a wrong uri is inserted

profile view delivers all projects
isAdmin is somehow broken (siderbar is evaluated correctly but all other places not...)

Drakulix:
Should all be resolved now

mharasic (dismissed):
Looks working

Rintel (approved):

PR #353, opened 2017-07-20
Profile shows email now and isAdmin works, I think the problem was th…

JamesComey:
…at at the logout, the loaded var was not set to false.
Oh yeah and the profile showas the email now!


Rintel (approved):

Drakulix:
Outdated and needs to be updated. Only the profile view change is relevant now that #351 is merged.

henderkes (changes requested):
Please update it to new master and remove unnecessary changes

henderkes:
Judging by the PR that was just merged, showing email on the profile page I guess this can be closed

PR #354, opened 2017-07-21
More advanced/flexible single search api

Drakulix:
Changes as requested by @mharasic 

~~Completely untested.~~

### DIFF
--- a/backend/__init__.py
+++ b/backend/__init__.py
@@ -151,14 +151,14 @@ class Notification(DB.EmbeddedDocument):
 
 class SavedSearch(EmbeddedDocument):
     saved_search_id = DB.ObjectIdField(default=ObjectId)
-    metadata = DB.StringField(max_length=4096)
-    query = DB.StringField(max_length=4096)
+    metadata = DB.StringField()
+    query = DB.StringField()
     count = DB.LongField()
 
     def to_dict(self):
         dic = {}
         dic['id'] = str(self.saved_search_id)
-        dic['meta'] = json.loads(str(self.metadata))
+        dic['query'] = json.loads(str(self.metadata))
         dic['count'] = self.count
         return dic
 
--- a/backend/api/search.py
+++ b/backend/api/search.py
@@ -27,7 +27,7 @@ def prepare_es_results(res):
         raise ApiException(str(ke), 400)
 
 
-@search.route('/api/projects/search/simple/', methods=['GET'])
+@search.route('/api/projects/search/simple/', methods=['POST'])
 @login_required
 def search_simple():
     """Search projects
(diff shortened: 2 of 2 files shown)


backend/__init__.py
```diff
@@ -151,14 +151,14 @@ def to_dict(self):
...
-    query = DB.StringField(max_length=4096)
+    metadata = DB.StringField()
+    query = DB.StringField()
     count = DB.LongField()
 
     def to_dict(self):
         dic = {}
         dic['id'] = str(self.saved_search_id)
-        dic['meta'] = json.loads(str(self.metadata))
+        dic['query'] = json.loads(str(self.metadata))
```
henderkes, line 161: any particular reason why we don't return 'meta' and 'query' seperately instead of returning metadata as 'query'?

backend/api/search.py
```diff
@@ -36,57 +36,41 @@ def search_simple():
...
-    if offset is None:
-        offset = 0
-    count = request.args.get('count', type=int)
-    if count is None:
-        count = 10
-    save = request.args.get('save', type=str)
+    query = request.get_json()
+
+    try:
+        searchString = query['searchString']
```
henderkes: should be search_string I think

backend/api/search.py
```diff
@@ -36,57 +36,41 @@ def search_simple():
...
+    except KeyError:
+        return make_response('Invalid json', 400)
 
-    # ^2 is boosting the attribute, *_is allowing wildcards to be used
     request_json = {
         'query': {
             'multi_match': {
-                'query': text,
+                'query': searchString,
```

backend/api/search.py
```diff
@@ -95,44 +79,67 @@ def search_avanced():
...
+            date_filter['range']['date_creation']['gte'] = date_from
+        if date_to:
+            date_filter['range']['date_creation']['lte'] = date_to
+
+        request_json['query']['filter'].append(date_filter)
+
+    if description:
+        request_json['query']['filter'].append({
+            'wildcard': { 'description': '*'+description+'*' }
+        })
```
henderkes: not really familiar with elasticsearch enough to comment on this properly, FE should test it before merge into master

backend/__init__.py
Drakulix: Yes I don't want to leak the ES query.

backend/api/search.py
```diff
@@ -95,44 +79,67 @@ def search_avanced():
...
+                'fields': [
+                    'tags^2',
+                    'title^2',
+                    'description',
+                ],
+            },
+            'filter': [
+                { 'term': { 'authors': authors } },
+                { 'term': { 'tags': tags } },
             },
```
mharasic: missing ] ...

backend/api/search.py
```diff
@@ -84,34 +84,51 @@ def search_avanced():
 
     try:
         search_string = query['searchString']
```
henderkes: do we really want "searchString" in the query? I feel like we should definitely be consistent with our other json key naming

backend/api/search.py
```diff
@@ -84,34 +84,51 @@ def search_avanced():
...
-        authors = query.get('authors', [])
-        tags = query.get('tags', [])
+        authors = query.get('authors')
+        tags = query.get('tags')
         status = query.get('status')
         date_from = query.get('date_from')
         date_to = query.get('date_to')
         description = query.get('description')
+        title = query.get('')
     except KeyError:
```
henderkes: since we use dict.get() everywhere catching KeyErrors won't work since get just returns None as default if the key doesn't exist. I think all the assignments should work with the __iter__ call instead (query['authors'] etc.)

Also title = query['']? Empty key?

henderkes (dismissed):

backend/api/search.py
Drakulix: this is consistent

backend/api/search.py
Drakulix: Yes thats the purpose, only searchString is required, so I don't want those to fire any key errors. And yes, I forget to add the title.

henderkes: Then just get rid of exception handling, get doesn't raise KeyErrors

Drakulix: But I do exception handling for searchString, because that is actually required, as written.

mharasic (dismissed):

PR #367, opened 2017-07-23
[FE] Fix redirect to discovery on reload

### LINKED ISSUE #363: Refresh of page leads to discovy

Drakulix:
... and fix broken site_path on the way.

The Switch only renders the first matching route hit. this avoids the redirect to '/' on reload, if a valid route would already been rendered with the set url.

Fix #363


mharasic (approved):
working

PR #373, opened 2017-07-24
[BE] test notifications with comment

henderkes:


backend/__init__.py
```diff
@@ -186,6 +186,7 @@ def to_dict(self):
         # dic['password'] = str(self.password)
         dic['active'] = str(self.active)
         dic['bio'] = str(self.bio)
+        dic['notifications'] = [noti.to_dict() for noti in self.notifications]  # TODO: do we want?
```
henderkes: Do we really want this? This means every user can check everyones notifications.

mharasic (approved):
Sounds legit

PR #376, opened 2017-07-24
Frontend/feature/admin for users

mharasic:


henderkes (dismissed):

henderkes (approved):

web/src/components/common/Backend.jsx
```diff
@@ -440,11 +440,11 @@ class Backend {
         return this.getJson('/api/users');
     }
 
-    setUserRoles(mail, roles){
-      return Promise.resolve().then(alert( " TODO  PLEASE CODE "+ mail + " " + roles))
+    setUserRoles(userInf){
```
Drakulix: Not a big fan of using the whole user object here instead of crafting it, like in the updateProfile call. This is quite inconsistent.

mharasic: changes.... but  everything has to be send ...

PR #389, opened 2017-07-24
Archive De-Archive now in table... Needs nicer Styling

mharasic:


henderkes (approved):

web/src/components/common/DataTable.jsx
```diff
@@ -351,42 +351,57 @@ export default class BookmarksTable extends Component {
         }
       })
     }
-    if(this.props.columns.indexOf("archive") !== -1){
+    if(this.props.columns.indexOf("dearchive") !== -1){
       columns.push({
-        Header: 'Archive',
+        Header: 'Unarchive',
```
henderkes: I think we should call it either Dearchive or Unarchive, but not sometimes the first and other times the other

henderkes (changes requested):

web/src/components/common/DataTable.jsx
henderkes: This is the case at multiple places btw, only marked it here

henderkes (dismissed):

web/src/components/common/Backend.jsx
```diff
@@ -441,20 +441,13 @@ class Backend {
...
     async setUserRoles(mail, firstName, lastName, bio, roles){
-      if (await this.putJson('/api/users', {
+      return this.putJson('/api/users', {
         'email': mail,
         'first_name': firstName,
         'last_name': lastName,
         'bio': bio,
         'roles' : roles
-      })) {
-        if (this.mail == mail) {
```
Drakulix: Why did you remove this check? What if I make myself from an admin to a normal user? Every profile update should always trigger a refresh of the profile, if it is mine.

Drakulix (dismissed):

Drakulix (approved):

PR #397, opened 2017-07-25
added user role update test, implement last admin check in update end…

henderkes:
…point

### DIFF
--- a/backend/api/users.py
+++ b/backend/api/users.py
@@ -105,9 +105,7 @@ def update_user():
                 res.last_name != user.get('last_name', res.last_name):
             res.first_name = user.get('first_name', res.first_name)
             res.last_name = user.get('last_name', res.last_name)
-            firstname = res.get('first_name', "")
-            lastname = res.get('last_name', "")
-            newname = firstname + " " if firstname else "" + lastname
+            newname = res['first_name'] + " " if res['first_name'] else "" + res['last_name']
             g.projects.update_many({'authors.email': user['email']},
                                    {'$set': {'authors.$.name': newname}})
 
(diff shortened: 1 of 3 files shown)


backend/api/users.py
```diff
@@ -112,18 +113,24 @@ def update_user():
...
+
         if 'roles' in user:
             if 'admin' in user['roles'] and not current_user.has_role('admin'):
                 raise ApiException("Current user has no permission to assign admin role.", 403)
-            res.roles = [g.user_datastore.find_or_create_role(role) for role in user['roles']]
+            if 'admin' not in user['roles'] and res.has_role('admin'):
+                for usr in g.user_datastore.user_model.objects:
+                    if usr.has_role('admin') and usr['email'] != user['email']:
+                        res.roles = [g.user_datastore.find_or_create_role(role)
+                                     for role in user['roles']]
```
Drakulix, line 122: What do you use `usr` for? I just see, that we are assigning res.role multiple times. Shouldn't we do it just once, if the current_user is allowed to edit res/user?

tests/api/test_users.py
```diff
@@ -14,11 +14,19 @@ def test_update_usernames(self, flask_api_url, enter_users, session, enter_data_
                                            document_id[0], json=document)
         print(put_project_response.text)
         assert put_project_response.status_code == 200
-        user['fist_name'] = "Dr. Dagobert"
```
Drakulix, line 17: Not sure if I liked the typo more. :joy:

backend/api/users.py
henderkes: this is to prevent taking the admin role away from the last admin, but you are right, it should only do that once. fixed with a break now

Drakulix: For loops do support else statements in python. I would like a nice error message in that case then, instead of a silent discard of the requested change.

henderkes: ops, comment wasn't posted in the review page: already done

Drakulix (dismissed):
In case travis is fine with this, I see nothing missing! :+1:

henderkes:
Yeah I accidentally used .get method on a User object, it's fixed now

Drakulix (approved):
Lets try this again.

PR #423, opened 2017-08-04
Dropped ES.. Looks like most working

mharasic:


backend/api/search.py
```diff
@@ -19,10 +18,8 @@
 def prepare_es_results(res):
     """ Internal function to prepare ElasticSearch search results to send as json.
     """
+    projects = [x for x in res[:]]
```
henderkes: list(res) is easier

backend/api/search.py
```diff
@@ -36,7 +33,7 @@ def prepare_es_results(res):
 def prepare_es_query(query):
     """ Internal function to prepare the ElasticSearch search query from a given json
     """
-    search_string = query.get('searchString', '')
+    search_string = query.get('searchString')
```
henderkes: this needs to be checked for None later now

backend/api/search.py
```diff
@@ -131,10 +76,10 @@ def search_es():
     query = request.get_json()
     request_json = prepare_es_query(query)
 
-    try:
-        projects = prepare_es_results(g.es.search(index="knexdb", body=request_json))
-    except RequestError as reqerr:
-        raise ApiException(str(reqerr), 400)
+    projects = g.projects.find(request_json, {'comments': 0})
+    projects = [x for x in projects[:]]
```

backend/requirements.txt
```diff
@@ -1,6 +1,4 @@
 flask
-pymongo
```
DubbleClick: pymongo must stay

backend/api/search.py
```diff
@@ -185,7 +130,7 @@ def query_saved_search(id):
                 query = json.loads(cursearch['query'])
                 query['size'] = count
                 query['from'] = offset
-                projects = prepare_es_results(g.es.search(index="knexdb", body=query))
+                projects = prepare_es_results(g.projects.find(request_json, {'comments': 0}))
```
DubbleClick: prepare_search results method is no longer needed

henderkes (dismissed):
okay I'll work on this tomorrow, still needs a ton of changes

backend/api/search.py
mharasic: It is still needed somehow.... It transforms it, but I will refactor

backend/api/search.py
mharasic: Nope... Indexes are smart enough to omit empty strings :)

henderkes: this will return None now instead of an empty string

henderkes: well... it's not used anywhere, so why bother getting it in the first place?

mharasic: It should go over the indixes, but can work with an empty string

henderkes (dismissed):

henderkes (approved):

PR #426, opened 2017-08-05
AVATAR !!!!

mharasic:


henderkes:
okay we definitely shouldn't save the images in the filesystem but stream the generated image into the database directly instead, I might have a look at it tomorrow

mharasic:
Small ugly workaround... the images are written to the FS and then deleted afterwards... but for sure it will work with streams

henderkes (dismissed):

henderkes (approved):
rebased to master

PR #427, opened 2017-08-06
BE Refactor

mharasic:


backend/api/helper/permissions.py
```diff
@@ -0,0 +1,16 @@
+from flask_security import current_user
+
+
+def current_user_has_permission_to_change(entry) -> bool:
```
henderkes, line 4: I dislike this very much compared to the is_permitted method before. The previous method call was even shorter due to the name and more versatile.

henderkes: Also if we already have a permission help class, we should also put the global ADMIN_PERMISSION in here. (or well, delete it since we don't use it currently)

backend/api/users.py
```diff
@@ -4,38 +4,27 @@
 
 import os
 import sys
-import json
 import base64
+import json
+
```
henderkes: pep8

henderkes (changes requested):
I guess it does somewhat make sense, but I think we should also change the endpoints then. /api/**users/**bookmarks/... does not make sense for the blueprint Bookmarks.

backend/api/helper/permissions.py
mharasic: It was more versatily but permissions are allways checked against the current user

mharasic: Yep... Admin Check should move in here

henderkes: yes I'm aware of that, but I'd rather have an extra wrapper for is_permitted called cur_permitted or something then. I can definitely see use for the is_permitted method if we add more user management stuff for the admin panel, or requesting project edits from a permitted user

mharasic: We use Admin-require in projects.py

mharasic: Yes! You are right for this

mharasic:
WIP

henderkes:
I still really don't agree with this method name for the permission check, I will work something out in two days or so, very busy currently. Also please don't merge master branch into this, I'll rebase at the end.

PR #430, opened 2017-08-07
Notification Rework

mharasic:
WIP


backend/api/helper/uploader.py
```diff
@@ -37,15 +37,9 @@ def save_manifest_to_db(manifest):
                     entry['archived'] = False
                 entry['authors'] = sorted(list(set(entry['authors'])))
                 entry['tags'] = sorted(entry['tags'])
-                g.projects.insert(entry)
+                projects.append(entry)
```
henderkes, line 40: erm, the uploaded projects will no longer be added to the database?

backend/__init__.py
```diff
@@ -116,21 +123,6 @@ class Role(DB.Document, RoleMixin):
     description = DB.StringField(max_length=255)
 
 
-class Notification(DB.EmbeddedDocument):
```
henderkes, line 119: ?

backend/api/notifications.py
```diff
@@ -0,0 +1,92 @@
...
+import uuid
+
+
+notifications = Blueprint('api_notifications', __name__)
+
+
+def add_notification(creator, userlist, project_id, operation, reason='', saved_search_id=''):
+    date = time.strftime("%Y-%m-%d %H:%M:%S")
+    for user in userlist:
+        if user not in (creator):
```
henderkes: this is a poor equality check

backend/api/notifications.py
```diff
@@ -0,0 +1,92 @@
...
+            g.notifications.insert_one({
+                '_id': str(uuid.uuid4()),
+                'creator': creator,
+                'user_id': user,
```
henderkes, line 19: do you think this is a good idea? notifications are pretty much made to be stored as EmbeddedDocuments.

henderkes (commented):
I don't think this is a good idea all around to be honest. Splitting the files into different blueprints is great, but I'm not a fan of the rest

backend/api/helper/uploader.py
mharasic: They are inserted in  projects.py

backend/api/notifications.py
mharasic: Feel free to make them as Embedded documents... Using pymongo only for convenience

backend/__init__.py
mharasic: USing pymongo !?!

backend/api/notifications.py
henderkes: They already were, that's why I'm confused you put them in their own collection now, this just seems confusing to me

backend/api/projects.py
```diff
@@ -46,27 +35,40 @@ def add_projects():
...
             else:
                 raise ApiException("Wrong content header" +
                                    "and no files attached", 400)
-
-            g.rerun_saved_searches()
-            return jsonify(return_ids)
+            project = projects[0]
+            if 'comments' not in project:
+                project['comments'] = []
+            g.projects.insert(project)
```
henderkes: only the first is saved, this seems like a huge mess

backend/__init__.py
```diff
@@ -273,9 +274,10 @@ def handle_invalid_usage(error):
 def handle_insufficient_permission(error):
     """ Handler for insufficient permission to access a method.
         This is not the error handler for insufficient permission to update
-        a project or user.
+        a project or user. It delivers a 404 as users should not see endpoints,
+        where they don't have permissions to.
```
henderkes, line 278: this handler is for cases where the user knows the endpoint but doesn't have permission to use it.
the other errorhandler is for cases where the user should not be aware of the endpoint

mharasic: Thats the 404 I'm not admin handler.... Strange

backend/api/projects.py
mharasic: Fixed..... Was trappend in a different understanding of code.... In that actual case we have unreachable code which was planned to work dirrently...

PR #433, opened 2017-08-16
Notification Setting implemented

mharasic:


backend/api/notifications.py
```diff
@@ -14,17 +16,22 @@ def add_notification(creator, userlist, operation, project_id='',
...
-                'project_id': project_id,
-                'operation': operation,
-                'date': date,
-                'reason': reason,
-                'active': 'true',
-                'saved_search_id': saved_search_id
-            })
+            user_object = g.user_datastore.get_user(user)
+            if user_object:
+                dict = user_object.to_dict()
```
henderkes, line 21: do not override default python names please, call it mydict or whatever

backend/api/notifications.py
```diff
@@ -14,17 +16,22 @@ def add_notification(creator, userlist, operation, project_id='',
...
+                notifications_settings = dict['notifications_settings']
+                if notifications_settings[operation] == 'true':
```
henderkes: I feel like we should use actual booleans here and not strings 'true' or 'false'. Also should use notification_settings.get(operation, False) to avoid crashes if operation is not a (valid) key in the dict.

henderkes (changes requested):

backend/api/notifications.py
mharasic: It is the function someOne created back in these days for the user

henderkes: still has to be changed, this is not pep conform, I don't know why the autocheck doesn't see this

mharasic: It is a class method of User.... should I rename it?

mharasic: Not working as supposed to be.... either you should rename it or it will be in the master nevertheless... that "issue" is at several places

